Issue #4, opened 2016-06-30
OTP versions supported officialy?


NobbZ:
I've seen in the README.md that there is only “official” support for 18.x because thats what you use @j14159. Personally I have to handle multiple versions on my system and swap back and forth all the time, and as such like to try to maintain compatibility in a given range. I already have an internal project which I have to keep compatible from 16B3 to current, but I do hope beeing able to drop 16 before years end.

Also I am trying to set up travis on my fork and wanted to know which versions of erlang I shall test against.

The current version does compile against the latest minor release of 17, 18 and 19, where 17 fails during compilation phase and 18 and 19 both pass all tests, but fail in dialyzer phase. For 17 there weren't even a dialyzer run, because of known trouble with modules generated from `xrl` and `yrl` files before OTP 18.

I can reproduce the exact same behaviour on my local system.

From what I have observed on my system and at travis, I'd guess supported range for now should be 18 and 19, while the tests should be run on 18.2, 18.3 and 19.0 at least, dropping 18.2 as soon 19.1 has been released. But I will follow what ever you suggest here.

I will do a WIP-PR in a couple of minutes.

j14159:
@NobbZ thanks for looking into this!  Iirc the dialyzer errors are due to a couple of things I should simply add to an ignore file (bad debugging helpers and a function generated by leex).  I think the range of versions you suggest (18.2, 18.3, and 19) are fine and travis integration is a good idea.

I had not used `rebar3 xref` before but it looks like it's throwing warnings on a lot of functions from the modules generated by leex and yecc.  Easy way to handle this or skip xref?  I'm open to suggestions.

NobbZ:
I already looked into this, and I can configure them easily away, at least the ones I can identify as leex and yecc related. Maybe one should take a deeper look into the remaining ones?

I'm only online by mobile right now, so I can't look as deep as I wish, but I can do that deeper look when I am back home in about 4 hours.

Also as far as I can remember there is a dialyzer complaint in 19, about impossible matches. I am not sure why that one doesn't happen in 18… Might be related to the dialyzer changes regarding maps.

Also in the meantime another question has come to my mind… Warnnings as Warnings or Warnings as Errors? There are a lot of unused variable warnings to fix… But they should get fixed either way ;)

j14159:
[...]

PR #5, opened 2016-06-30
[WIP] Add travis.yml for continuos integration [WIP]

NobbZ:
This particular branch is tested by travis for now on my own account. Logs can be seen at [travis-ci.org/NobbZ/mlfe](https://travis-ci.org/NobbZ/mlfe).

Before merging this, there are some open questions, which I want to discuss either here or in separate tickets.

- [x] Targetted OTP-Versions (#4)
- [x] Shall `rebar3 xref` integrated into the build?
- [ ] Shall there be an icon showing current travis-build status in the README.md?
- [ ] Warnings as warnings or warnings as errors?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: erlang
 otp_release:
   - 19.0
   - 18.3
-  - 17.5
+  - 18.2
 
 env:
   global:
@@ -24,5 +24,5 @@ before_install:
 before_script: [] # Need to do this, or travis will use rebar2!
 
 script:
-  - if [ "$TRAVIS_OTP_RELEASE" != "17.5" ]; then rebar3 dialyzer; fi # dialyzer before OTP 18 will fail on modules generated from XRL or YRL files.
+  - rebar3 dialyzer
   - rebar3 eunit
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+language: erlang
+otp_release:
+  - 19.0
+  - 18.3
+  - 17.5
+
+env:
+  global:
+    - REBAR_VSN=3.2.0
+    - PATH=./rebar3-$REBAR_VSN:$PATH
+
+cache:
+  directories:
+    - .rebar3-$REBAR_VSN
+    - /home/travis/.cache/rebar3
+
+before_install:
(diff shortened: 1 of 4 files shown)


NobbZ:
I added xref to the build and added pragmas to ignore some functions for xref as well as dialyzer, because we do not have control about them in generated code. Also I ignored `mlfe:compile/1` as it is obvious to me that it is some kind of entry point from extarnal OTP-Apps.

There are still some xref warnings open where I am not sure iff they are allowed to be silently ignored or properly handled.

Also I took a look at the warnings dialyzer spits out when using OTP 19. I have to conclude that dialyzer is right, but do not ask, why dialyzer from 18 does not catch them.

These warnings are about trying to match the `tail` of a `#mlfe_cons{}`, the `tail`-field itself is allowed to be either an `#mlfe_cons{}` or `{nil, integer()}` (destructured types a bit for making it easier to follow along). But in fact, the `case`-`of` where the warnings are emitted trys to match against `{symbol, L, _}`, `#mlfe_apply{}`, and `NonList`.

* `{symbol, L, _}` is not in `#mlfe_cons{} | {nil, integer()}`,
* `#mlfe_apply{}` is not in `#mlfe_cons{} | {nil, integer()}`,
* `NonList` can't match because former matches cover `#mlfe_cons{} | {nil, integer()}` completely.

I do not know if this is a misspecification or some oversight in the case. So I do not know If I had to repair by expanding the typespec or by removing code from the `case`-`of`.

So the remaining items remain yours to fix.

I you want me to, I will do another commit which allows 19 to fail for now until the typespec stuff can be further evaluated.

j14159:
I think `tail`'s type in `#mlfe_cons{}` can be relaxed to `mlfe_expression()` for simplicity here.  Cons'ing to a symbol tail is permitted in that method since the symbol may very well type to a list itself of course.  The catch-all `NonList` case should be ok here and correct, R18's dialyzer reports no errors with this change although I'm not certain I should take that as a definitive answer given what you're seeing above ;)

It seems to me that the change to `mlfe_expression()` for the tail and letting the typer rule out cons'ing onto anything that _can't_ be typed to a list is correct anyhow.  I'll merge this PR and close this ticket, thanks very much for the contribution!  If you've strong opinions that differ on the list changes I'm suggesting, happy to discuss them under a separate issue.  I'll check the other xref errors and squash what makes sense.

Thanks again!

PR #12, opened 2016-08-04
Fix unary minus and plus

danabr:


danabr:
Let me know if you want the tests in another module than basic_math.mlfe.

j14159:
Thanks for this!

PR #22, opened 2016-08-12
Cleanup

tuncer:
@j14159, here's the result of cleanups (consistent formatting, correct comment syntax, whitespace errors, etc.), as discussed.


tuncer:
There are some very, very, deeply nested functions that could benefit from code extraction.

j14159:
@tuncer thanks for this!  Agreed, `mlfe_typer.erl` especially would benefit from an overhaul although with the uneven application of generalization and processes-as-reference-cells I'm tempted to reimplement a lot of it tbh.

PR #24, opened 2016-08-13
Fix unification of polymorphic adts.

### LINKED ISSUE #23: Type checker does not check polymorphic user defined types
The following module should not pass type checking, but it does:

````
module tree

export height/1, fail/1

type tree 'a = Leaf | Node (tree, 'a, tree)

height t = 
  match t with
    Leaf -> 0
  | Node (l, _, r) -> 1 + (max (height l) (height r)) 

max a b = 
  match (a > b) with
    true -> a
  | false -> b

fail () = height 1
````

danabr:
Closes #23.


j14159:
Awesome, thanks!

Issue #20, opened 2016-08-12
User defined types should be parameterizable with builtin types


danabr:
Consider this example:
````
module user_defined_types

type proplist 'k 'v = list ('k, 'v)

type optlist 'v = proplist atom 'v
````

This fails with:
````erlang
{error, {badmatch, {error, {6,mlfe_parser, ["syntax error before: ",["\"atom\""]]}}}
````

The parser expects a user defined type to only take type variables. See [poly_type](https://github.com/j14159/mlfe/blob/master/src/mlfe_parser.yrl#L102) in mlfe_paser.yrl.

mlfe_typer also has this assumption.

j14159:
Ah yes - looks like the parser only permits the base polymorphic types to do this.  Should be a pretty simple fix with error checking/validation almost entirely occurring in `mlfe_typer.erl`.  Nice find, thanks!

j14159:
Oh, actually, maybe we just need `poly_type` to take a list of type expressions instead of type vars.

danabr:
I also think we should not define `base_type`, `base_list`, etc in the lexer. It prevents us to use these for identifiers: This code does not compile for example:

````
map f l = 
    match l with
        [] -> []
    | h :: t ->
        (f h) :: (map f t)
````

j14159:
Sure, I just didn't want them to be replaceable as types.

j14159:
@danabr let me know if you'd like to take a crack at this - if not I'll start looking at it this weekend.  Just don't want to duplicate effort :)

danabr:
I'm playing with it now. I'm not sure I will be able to make anything mergable, though. I'll let you know tomorrow.

j14159:
No worries.  The typer is...convoluted at the moment, happy to help out.  I'll leave this issue alone in favour of working through records some more (no branch yet, just reading and note-taking).

danabr:
I don't think I'm going to make it. I can get the parser to work somewhat as I want, but I cannot wrap my head around the typer. See [this branch] (https://github.com/j14159/mlfe/compare/master...danabr:fix/user-defined-types-parameterized-with-builtin-types) for a failing test case.

danabr:
Seems like the issues I ran into were caused by the issue described in #23. After applying only the parser changes on top of #24 my original example seems to work. I'll try to get around to cleaning it up and creating a PR.

danabr:
Closed by #27.

PR #27, opened 2016-08-15
Fix/user defined types parameterized with builtin types

danabr:


danabr:
This is actually more parsing related than typing related. Should I move the test to the `mlfe` module instead?

danabr:
Closes #20.

src/mlfe_parser.yrl
```diff
@@ -96,14 +96,14 @@ type_import -> use module_fun:
 
 module_def -> module atom : {module, '$1'}.
 
-type_vars -> type_var : ['$1'].
-type_vars -> type_var type_vars : ['$1'|'$2'].
+type_vars -> type_expr : ['$1'].
```
j14159: Should we maybe change from `type_var` to something else like `type_expressions` or something along those lines?  Minor point but might alleviate confusion later :)

j14159:
I'm OK with the test in either place as it's a parsing fix for a typing feature.  Wherever you're most comfortable with it.

j14159:
Thanks for this!

PR #28, opened 2016-08-15
Allow modules that do not export any functions

danabr:
I'm not sure if this makes sense. In any case we should not crash. If we don't want type-only modules, we should return an error instead.

### DIFF
--- a/src/mlfe_typer.erl
+++ b/src/mlfe_typer.erl
@@ -1011,9 +1011,9 @@ typ_module(#mlfe_module{functions=Fs,
 
                     case type_module_tests(Tests, Env3, ok, FunRes) of
                         {error, _} = Err        ->
-                          Err;
+                            Err;
                         Funs when is_list(Funs) ->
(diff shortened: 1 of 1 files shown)


src/mlfe_typer.erl
```diff
@@ -1010,8 +1010,10 @@ typ_module(#mlfe_module{functions=Fs,
                     FunRes = typ_module_funs(Fs, Env3, []),
 
                     case type_module_tests(Tests, Env3, ok, FunRes) of
-                        {error, _} = E -> E;
-                        [_|_] = Funs   -> {ok, M#mlfe_module{functions=Funs}}
+                        {error, _} = Err        ->
+                          Err;
```
j14159: Minor nit-pick on indentation here - are you using an emacs or vim mode for Erlang?

danabr: vim. It is just that my default indentation is 2 spaces. I better start using something like erlfmt before I push.

PR #29, opened 2016-08-16
Add support for returning functions from functions

danabr:


j14159:
Nice find, thanks!  Similar indentation nit-pick to #28.

PR #30, opened 2016-08-17
Fixes for applications of functions with polymorphic return types.

### LINKED ISSUE #25: Function gets constrained by callers
Consider this example:

````
module list_tests

is_empty l =
  match l with
    [] -> true
  | _ :: _ -> false

a () = is_empty []

b () = is_empty [:ok]

c () = is_empty [1]
````

This fails unexpectedly with `{error,{cannot_unify,list_tests,12,t_atom,t_int}}`.

The fact that we applied an atom to is_empty, should not specialize the function to only act on atoms.

danabr:
Closes #25.

### DIFF
--- a/src/mlfe_typer.erl
+++ b/src/mlfe_typer.erl
@@ -115,6 +115,20 @@ copy_cell(Cell, RefMap) ->
                 Existing ->
                     {Existing, RefMap}
             end;
+        {t_list, C} ->
+            {C2, Map2} = copy_cell(C, RefMap),
+            {new_cell({t_list, C2}), Map2};
+        {t_map, K, V} ->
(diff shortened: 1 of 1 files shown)


danabr:
I'm quite sure we should add clauses for tuples, pids and receivers as well. However, I'm not sure I've added the fixes to the correct place. So I'm fixing just ADTs, lists and maps now.

j14159:
Good catch, we'll definitely need to do this for any polymorphic type so you're right about tuples, pids, and receivers.  I'll merge this but I think #25 should stay open and I'll leave a comment to that effect on the issue.

j14159:
I should say:  for any type containing more reference cells, not necessarily just polymorphic :)

PR #31, opened 2016-08-18
Avoid leaking processes while type checking.

danabr:
This is achieved by performing the type checking in a separate
process, and linking all reference cells to that process.

### DIFF
--- a/src/mlfe_ast.hrl
+++ b/src/mlfe_ast.hrl
@@ -77,7 +77,7 @@
              | t_map()
              | t_tuple()
              | t_clause()
-             | typer:t_cell().  % a reference cell for a type.
+             | mlfe_typer:t_cell().  % a reference cell for a type.
 
 %%% ## MLFE AST Nodes
 
--- a/src/mlfe_typer.erl
+++ b/src/mlfe_typer.erl
@@ -3206,6 +3206,16 @@ no_process_leak_test() ->
     {ok, _, _, M} = mlfe_ast_gen:parse_module(0, Code),
     ProcessesBefore = length(erlang:processes()),
     ?assertMatch({ok, _}, type_modules([M])),
-    ProcessesAfter = length(erlang:processes()),
+    ProcessesAfter = wait_for_processes_to_die(ProcessesBefore, 10),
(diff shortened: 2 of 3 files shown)


src/mlfe_ast.hrl
```diff
@@ -77,7 +77,7 @@
              | t_map()
              | t_tuple()
              | t_clause()
-             | typer:t_cell().  % a reference cell for a type.
+             | mlfe_typer:t_cell().  % a reference cell for a type.
```
danabr: This is the line introducing the dialyzer errors. You get the same thing on master if you change the same line. Will try to fix the warnings in a separate branch.

danabr:
Depends on dialyzer fixes in #32.

danabr:
Hmm, it seems that the test I added is flaky. It seems like that the VM does not always reap the processes fast enough.

Shall I remove the test?

j14159:
Would sleeping for 100ms or so be bad form do you think?

tuncer:
If you want to wait, I'd suggest a loop that slightly increases the delay on each check and gives up after N tries.

src/mlfe_typer.erl
```diff
@@ -980,9 +984,35 @@ retrieve_type(SM, M, T, [#mlfe_module{name=M, types=Ts}|Rem]) ->
...
+                                       {error, term()}.
+type_modules(Mods) ->
+    {Pid, Monitor} = erlang:spawn_monitor(fun() ->
+        try type_modules(Mods, new_env(Mods), []) of
+            Res -> exit(Res)
+        catch
+            E:T ->
+                io:format("mlfe_typer:type_modules/2 crashed with ~p:~p~n"
+                          "Stacktrace: ~p~n", [E, T, erlang:get_stacktrace()]),
+                exit({error, "mlfe_typer:type_modules/2 crashed"})
```
tuncer: I suggest to insert a linebreak before printing the potentially large terms. That makes it more likely to fit into constraints and will also make it easier to copy paste since it starts at column zero. Also, in case you want the whole terms to be printed, you might want to reconsider the use of `~p`.

PR #32, opened 2016-08-18
Fix type declarations.

danabr:
The types should really be defined in `erl` files, and not in `hrl` files, but I leave fixing that for later.

This also shows the danger of "unknown types". Unfortunately, it doesn't seem like rebar3 warns about them.

### DIFF
--- a/src/mlfe_ast.hrl
+++ b/src/mlfe_ast.hrl
@@ -38,9 +38,9 @@
 %% allowed to be sent messages.
 -type t_pid() :: {t_pid, typ()}.
 
--type t_cons() :: {t_cons, typ(), t_list()}.
--type t_nil() :: t_nil.
--type t_list() :: t_cons() | t_nil().
+-type t_receiver() :: {t_receiver, typ(), typ()}.
+
+-type t_list() :: {t_list, typ()}.
 
 -type t_map() :: {t_map, typ(), typ()}.
 
@@ -77,7 +77,10 @@
              | t_map()
              | t_tuple()
              | t_clause()
-             | typer:t_cell().  % a reference cell for a type.
+             | t_pid()
+             | t_receiver()
+             | mlfe_type()
+             | mlfe_typer:t_cell().  % a reference cell for a type.
 
 %%% ## MLFE AST Nodes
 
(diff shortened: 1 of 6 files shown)


danabr:
Builds and passes dialyzer and all tests locally, but dialyzers gets killed on TravicCI, probably due to excessive memory usage.

@j14159: Could you try re-trigger the build?

j14159:
@danabr retriggering fails as well, looks like dialyzer is exiting with error code `137`, appears to be getting killed as you suggested.

j14159:
Limiting parallel tasks to one doesn't seem to alleviate this either.

src/mlfe_ast.hrl
```diff
@@ -77,7 +77,10 @@
              | t_map()
              | t_tuple()
              | t_clause()
-             | typer:t_cell().  % a reference cell for a type.
+             | t_pid()
+             | t_receiver()
+             | mlfe_type()
```
danabr: This is the line that causes dialyzer to run out of memory.

danabr:
Dialyzer now passes on OTP 18, but fails on OTP 19. I think the dialyzer spec for `compile:forms` in OTP 19 is wrong (it requires a list of forms, but accepts a single module form as well). Tried to check what LFE does, but they don't build on OTP 19 yet.

danabr:
As I suspected, this was a bug in OTP. Fixed in [OTP master](https://github.com/erlang/otp/commit/9aabc64783ee1a1fbcd5b80ea03f9ef12807cb92), but not in any of the tagged OTP 19 versions.

@j14159: Shall we disable building on OTP 19 for now?

j14159:
I hate to disable it but that might be best.  Out of curiosity, have you tried moving the type definition for `t_cell()` into the `.hrl` file?  Curious if that helps at all though would be surprised if so.

danabr:
Disabled the build on OTP 19. We can enable it once they release 19.1, where the fix to the `compiler:forms/2` spec is included.

j14159:
Mind adding something to the README mentioning the R19 issue?  I think I mention using R19 specifically there.

danabr:
I'll update the README, and try to just skip dialyzer for OTP 19.0... https://travis-ci.org/danabr/mlfe/builds/153654314

j14159:
Sounds good, thanks!

danabr:
It worked :)

PR #35, opened 2016-08-24
Do not allow type names on the LHS of type declarations.

danabr:
The fix for #27 accidentally introduced the ability to specify nonsense type declarations, such as:

````
module garbage

type garbage int = GarbageIn  | GarbageOut

````

As the updated test case show, I even fell for it myself!


j14159:
Nice catch!  Apologies for the delayed response, have been travelling.

Would it make sense to add a test checking for the expected failure to prevent a regression?

PR #39, opened 2016-09-09
Fix error propagation if typ_of fails on a tuple member

danabr:


danabr:
In general I think it would be much better to propagate errors via the `erlang:throw` and have a catch very high up in the chain.

j14159:
Agreed.  Iirc I put the propagation stuff in due to the lack of line numbers being passed through but that's less of an issue now.  Moving to `erlang:throw` would certainly slim down a lot of the typer code :)  If not immediately, definitely something to consider during a rewrite to use something like ETS for reference cells rather than processes (been bugging me for a while).

PR #58, opened 2016-12-01
Add basic infix functions

lepoetemaudit:
This adds the ability to add simple infix functions beyond the current builtins. Currently they must be at least two chars in length and cannot override any of the builtins.

Resolves https://github.com/j14159/mlfe/issues/55


j14159:
It would be great to add a couple of quick "happy path" tests to start, specifically:

- exercising the `|>` function in `test_files/simple_example.mlfe`, maybe this needs to go through a different proxy function in the module for Erlang to call it?  Not sure.
- a quick typer test to ensure a user-defined infix function types the way we want it to.

Thanks again!

lepoetemaudit:
Thanks for the pointers, I'll see what I can come up with.

PR #78, opened 2016-12-19
Prefix generated modules with `alpaca_`

### LINKED ISSUE #75: Consider prefixing compiled Alpaca modules
When Elixir compiles modules, it prefixes the generated module name with 'Elixir.' ([more info here](http://elixir-lang.org/crash-course.html#adding-elixir-to-existing-erlang-programs)). This means that any modules compiled with Elixir don't risk clashing with Erlang ones, and it keeps the Elixir standard library nicely namespaced. From within Elixir code, you don't need to use the prefix.

I believe the same would be useful in Alpaca - perhaps automatically prefix every generated module with `alpaca_`. It means a bit of extra typing when calling from the Erlang side, but it also means we could give nice names to Alpaca standard library modules like `string` instead of having to import e.g. `alpaca_string` within Alpaca code.

lepoetemaudit:
Fixes #75. I haven't added any specific tests for this as the existing test suite causes this to be tested in as many useful ways as I could conceive.

We might also want to consider renaming the Alpaca source files included (which may become parts of the compiler and the stdlib) as they all currently have an `alpaca_` prefix, meaning that they get double prefixed once compiled.

### DIFF
--- a/src/alpaca.erl
+++ b/src/alpaca.erl
@@ -86,9 +86,10 @@ load_files(Filenames) ->
 compile_module(#alpaca_module{name=N}=Mod, Opts) ->
     {ok, Forms} = alpaca_codegen:gen(Mod, Opts),
     {ok, _, Bin} = compile:forms(Forms, [report, verbose, from_core]),
+    PrefixedName = list_to_atom("alpaca_" ++ atom_to_list(N)),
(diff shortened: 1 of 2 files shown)


j14159:
Looks good, thanks for this!

PR #81, opened 2016-12-27
Value literals

### LINKED ISSUE #80: Top level values (or nullary functions) are not usable at present
Currently, it is possible to define, but not use, zero-argument (nullary) functions. For example, the following throws a compilation error:

```elm
module example

x = 10

run () = x + x
```

Specifically:
```erlang
{cannot_unify,main,5,t_int,{t_arrow,[],<0.178.0>}}
```

In other words, it's failing the type check because `x` is compiled as a zero-arg function, whereas the function `+` expects only integers. However, this compiles:

```elm
module example2

x = 10

run () = x
```

When called from Erlang, `run ()` returns the zero arg function `x`, which can be invoked from Erlang to produce the value 10, but as far as I can tell but there is no way of getting the return value of `x` within Alpaca.

@j14159 has stated, and I agree, that nullary functions are not desirable and that it would be better if `x` is understood in this circumstance as a constant value. We discussed this on IRC and raised several issues:

1. When should the value be calculated? Compile time? Runtime (i.e. on module load?)
2. Where should the calculated values be stored? @j14159 suggested ETS, with the drawback that it would potentially be modifiable outside of the module
3. Would we allow side effects in values? OCaml, for example, allows this:
```ocaml 
let _ =
  print_string "Hello\n";;
```

lepoetemaudit:
Fixes #80. This PR causes zero-arg functions to be treated as constant values. This is achieved via the following:

- At the parser stage, zero-arg functions (at the top level) are checked to have a body containing a literal only (atom, string, chars, int, bool, float, record, map, list, binary, tuple, with the literal requirement recursing down to the constituent members of 'container' types like tuple, map etc.). This is to ensure referential transparency.
- In the typer, zero-arg functions are considered to be equivalent to their return type, i.e. `x = 10` is typed as `int`, which allows the use of these values in expressions.
- At the code generation stage, a reference to a zero-arg function is substituted for a call to it, meaning its return value is used rather than the value of the function itself.


yurrriq:
Without having read the code, this sounds great, thanks!

j14159:
Thanks for this @lepoetemaudit !

PR #83, opened 2016-12-29
Fix some whitespace issues

ypaq:
* fix whitespace regex to cover whitespace other than spaces
* fix block comment regex to cover more edge cases
* fix line comment regex
* add basic framework for property based testing using proper
* property test for compiling generated module declarations


j14159:
Thanks for this!

Issue #82, opened 2016-12-28
Double line breaks in expressions fail to compile


ypaq:
It would be desirable to be able to split code up with several line breaks. This is an example where double line breaks `\n\n` are significant and cause a compilation error.

```
alpaca:compile({text, [<<"module a \n\nf a = let add x = x + x in\n\nadd a a\n\n">>]}).
```

lepoetemaudit:
I checked OCaml's parser, which has a broadly similar syntax to Alpaca's - ML but non-whitespace significant (the above notwithstanding). In OCaml, top level phrases have to be terminated with `;;` to avoid the above issue, but they can be elided in certain cases, such as before an `open` or `let` - [some more info here.](https://ocaml.org/learn/tutorials/structure_of_ocaml_programs.html#Usingandomittingand)

With that in mind, I can think of three options:

1. We introduce `let` at the top level (and possibly optional terminator `;;`) and follow much the same parsing rules as OCaml.
2. We switch to whitespace significant rules à la Haskell, Elm and F#
3. We can be a bit less drastic than option 2 and simply require that top level function bodies are indented - they terminate either at EOF or the next non-empty line that begins at column 1.

I think 3 would solve this particular issue in a fairly non-intrusive way, although it might have other side-effects I haven't considered.

j14159:
I like options 1 and 2 a lot more than 3 to be honest.  3 feels like we're going in python's direction where right now we're just treating `\n\n` as punctuation, seems simpler to me.  How far away from #2 are we do you think?  Any big inconsistencies or horribleness?

I'm also ok with seriously considering introducing top-level `let`s and then sort of privileging `module`, `import`, `export`, `let` et al as signifying the major AST nodes for the parser.

To boil it down, my current opinion is:
- `let` is a little noisy but far from terrible.  Does seem one way of simplifying things with a bit of an invasive fix in the immediate term.
- I'm ok with significant whitespace if it's limited to `\n` being significant (it being treated as punctuation).
- there's going to be a code formatter.

I'm pretty open to changing everything but the formatter :)

lepoetemaudit:
[...]

PR #85, opened 2016-12-31
Add 'let' to top level definitions; infer breaks using top-level keywords

lepoetemaudit:
This PR represents a small shift in how we lex/parse statements at the top level.

Previously, all top level statements had to terminate with at least two new lines, i.e. `\n\n`. This meant encountering situations such as this: #82 

To remedy this, and following discussions with @j14159 and @ypaq on IRC, this proposed PR includes the following changes:

1. All whitespace significance is removed.
2. After the initial scanning, breaks are inserted into the token list by logically determining top level keywords. This has to be done with some notion of state in this style as 'type' and 'let' may have a different meaning depending on their context. This step can be found [here](https://github.com/alpaca-lang/alpaca/compare/master...lepoetemaudit:top-level-let-ignore-ws?expand=1#diff-5a99f1ea2b47a1485d1c6a88d20a180aR17).
3.  We now require 'let' before top level function / value definitions. It's noisier than before, but there isn't any way of otherwise distinguishing these from other top level constructs, so it was this or add more whitespace significance and 'offside' rules such as in Haskell/Elm. On the bright side, more OCaml code just became copy-pastable into Alpaca :)
4. The previous fix to sanitise groups of newlines is no longer required so it has been removed

I also made it aware of the up-coming 'import' keyword so hopefully integrating that will be smooth.

### DIFF
--- a/src/alpaca_ast_gen.erl
+++ b/src/alpaca_ast_gen.erl
@@ -741,13 +741,13 @@ defn_test_() ->
                        versions=[#alpaca_fun_version{
                                     args=[], 
                                     body={int, 1, 5}}]}},
-                   parse(alpaca_scanner:scan("x=5"))),
+                   parse(alpaca_scanner:scan("let x=5"))),
      ?_assertMatch(
         {ok, {error, non_literal_value, {symbol, 1, "x"}, 
                      {alpaca_apply,undefined,1,
                        {symbol,1,"sideEffectingFun"},
                        [{int,1,5}]}}},
-        parse(alpaca_scanner:scan("x=sideEffectingFun 5"))),
+        parse(alpaca_scanner:scan("let x=sideEffectingFun 5"))),
      ?_assertMatch(
         {ok, {error, non_literal_value, {symbol, 1, "x"},                  
                          {alpaca_record,2,1,false,
@@ -757,7 +757,7 @@ defn_test_() ->
                                   {alpaca_apply,undefined,1,
                                       {symbol,1,"sideEffectingFun"},
                                       [{int,1,5}]}}]}}},
-        parse(alpaca_scanner:scan("x={one = 10, two = (sideEffectingFun 5)}"))),        
+        parse(alpaca_scanner:scan("let x={one = 10, two = (sideEffectingFun 5)}"))),        
      ?_assertMatch(
         {ok, {error, non_literal_value, {symbol, 1, "x"}, 
                      {alpaca_cons,undefined,0,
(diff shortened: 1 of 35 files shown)


lepoetemaudit:
Seems to be getting stuck on the common test stage... Seems to be hanging for me locally unless it's just really slow?

j14159:
Travis-CI is killing the `ct` stage, could be time or memory limits.

j14159:
@lepoetemaudit the number of tests `proper` is running for `ct` is set to 1000 at the moment - maybe try cutting that in half?  Line 12 of `test/alpaca_SUITE.erl`.

lepoetemaudit:
@j14159 thanks, I actually tried setting it to 1 (!) locally and same effect, so I can only assume I broke it somehow. I'll investigate :)

lepoetemaudit:
Yes, I think that's it - I had a quick stab at adding the let but I'm not very familiar with quickcheck et al. Will figure it out and update in due course.

lepoetemaudit:
Fixed - it turns out removing the sanitization fix I put in a while for tabs in-between newlines caused the scanner to hang in some corner cases. I'm not entirely sure why, though.

j14159:
Thanks for this!  I know it's a bit of a slog to walk through fixing all of those tests.

Before I merge, do you think it's worth expanding the definition of `WS` in the lexer to include `\n` and just do away with tokenizing breaks entirely?  I'm not sure but it might elide the need for the cleanup before inferring breaks.  I'm easy either way.

yurrriq:
Maybe we can write some PropEr tests. I found them invaluable when writing life's docstring support.

yurrriq:
https://github.com/rvirding/lfe/blob/develop/test/prop_lfe_doc.erl

lepoetemaudit:
Amending all the lets was a nifty way of getting familiar with all the tests so it wasn't so bad :)

I think it would be better to fix the scan as you suggest, it's a bit of a smell otherwise. It's possible to cause infinite loops in Lecc with non-greedy matches which is what I think is happening here. I'll see what I can figure out.

@yurriq this whitespace issue was caught by proper, as added by @ypac - it's very clever stuff which I'm looking forward to getting more familiar with!

yurrriq:
> this whitespace issue was caught by proper

Oh, cool! I guess I missed that.

j14159:
Per discussion with @lepoetemaudit in IRC I'll merge this now with discussed scan fixes to come later.  Fixes #82

PR #89, opened 2017-01-04
Scanner fixes

lepoetemaudit:
This removes the sanitisation regex before scanning but implements its side-effect: coercing all input to lists, as Leex appears to choke otherwise. Also, Leex was hanging on unknown tokens such as apostrophe. I added a catch-all which fixes this.


j14159:
Thanks for this!

PR #92, opened 2017-01-13
Perform exhaustiveness checking of top level functions.

danabr:
Provides an option `warn_exhaustiveness` that will cause warnings to be
printed in case a top level function does not cover all the possible
patterns for its argument types.

Note that the check will be run regardless of whether the option is set
or not. The option only affects printing of warnings messages. This is
to ensure the checker evolves with the rest of the code.

No exhaustiveness checks are provided for maps. A true exhaustiveness
check would check that all combinations of presence/absence of the
possible keys for the key type match all the possible combinations of
the value type. This is clearly not of any practical value though.

A more pragmatic exhaustiveness check for maps would be to check value
coverage for all provided keys. However, this requires looking at the
provided patterns to find the covering patterns, which the current
algorithm does not support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ CodeAsAString}, [options, again])` for now.
 Supported options are:
 * `'test'` - This option will cause all tests in a module to be type checked and exported
     as functions that  [EUnit](http://erlang.org/doc/apps/eunit/chapter.html) should pick up.
-* `'warn_exhaustiveness'` - If provided, the compiler will print warnings regarding missed patterns in top level functions.
+* `{'warn_exhaustiveness', boolean()}` - If set to true (the default), the compiler will print warnings regarding missed patterns in top level functions.
 
 Errors from the compiler (e.g. type errors)
 are almost comically hostile to usability at the moment.  See the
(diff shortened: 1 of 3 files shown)


danabr:
I suggest checking out the branch and playing with it to see the actual printouts.

j14159:
I've got a mild preference for always printing/logging warnings with a flag to fail with errors on non-exhaustive matches.  Thoughts?

danabr:
I think I'd rather go with a generic  [warnings_as_errors] option.

Regarding always printing the warnings, that would work for me, since I usually prefer total functions. Others might not want to write the extra ` let my_fun other = error {assumption_violated, other}` clause. Perhaps being able to add annotations on the expression level would let you choose one a case by case basis.

Suggestion:
* Add warnings_as_errors option
* Add a nowarn_exhaustiveness option to disable exhaustiveness warnings
* Enable exhaustiveness warnings by default.

yurrriq:
:+1: to @danabr's suggestions.

ferd:
for what it's worth in terms of options design, I'd recommend `{warning_as_errors, boolean()}` as options as opposed to having variants like `warning_as_errors` + `nowarn_as_errors` for example, since the booleans are more easily overridable in tools like rebar3 (which alpaca uses) without needing to bake in special support when dealing with profiles. On/Off toggles work better than disjoint options influencing the same behaviour.

danabr:
I've changed to print warnings by default and slightly updated the documentation.

Implementing `{warnings_as_errors, boolean()}` would suit better to the generic compiler framework work that has been discussed elsewhere.

j14159:
Thanks for this!

PR #96, opened 2017-01-17
Support functions as ADT arguments.

danabr:


j14159:
Thanks for this!

danabr:
Having thought a bit more about this, I think the actual syntax we chose is misleading. In Haskell, OCaml, and Standard ML `int -> bool -> float` means "a function taking an int, and returning another function taking a bool and returning a float". In Alpaca, it means "a function taking an int and a bool, and returning a float".

I think we should pick another syntax to clarify that we actually have multi-arity functions. e.g. `[int,bool] -> float` or `(int, bool) -> float` or similar.

lpil:
I agree. Perhaps open an issue?

j14159:
@danabr makes sense.  The list syntax works for me if you're happy with it.

danabr:
Opened PR #100 to address this.

PR #100, opened 2017-01-18
Wrap parameter lists in types with '[]'.

danabr:
That is, the type of the following function is written `[int,int] -> int`,
not `int -> int -> int`:
````
foo a b = a + b + 1
````

This makes the arity of the function explicit.

### DIFF
--- a/src/alpaca_parser.yrl
+++ b/src/alpaca_parser.yrl
@@ -18,7 +18,7 @@ comment
 op infix
 const module_fun
 type poly_type poly_type_decl type_vars type_member type_members type_expr
-type_expressions type_tuple type_tuple_list
+type_expressions type_tuple comma_separated_type_list type_list
 type_apply
 type_import type_export types_to_export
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by j14159** [comment]
Thanks!

**Comment by lpil** [comment]
I feel this syntax would confused for a list of a or b

**Comment by danabr** [comment]
I agree it is not very beautiful. Feel free to come with other suggestions that might feel more natural. I just wanted to rectify the flaw of the initial notation before it spread to other parts of the code base.

**Comment by OvermindDL1** [comment]
Could use curly brackets?
```ocaml
val f : { int, int } -> int
let f x y = x + y
```

lpil:
Sounds good, assuming `{}` isn't used for any literals.

lpil:
Oh, wait, it's kinda similar to records.

OvermindDL1:
Hmm true, well you could always use a decorate list or so:
```ocaml
val f : ![ int, int ] -> int
let f x y = x + y
```

Single element tuples are rare though, perhaps a function definition inside one?
```ocaml
val f : (int -> int) -> int
let f x y = x + y
```
It would compose nicely, and if someone wants to pass a function definition in a 1-tuple then just:
```ocaml
val f : ( (int -> int) ) -> int
let f (cb) = cb 42
```

Though I'm still partial to an explicit type:
```ocaml
val f : ( int, int ) fn -> int
let f x y = x + y
```
Or even:
```ocaml
val f : ( int, int, int ) fn
let f x y = x + y
```
Where the first element in the tuple of the `tuple fn` type is the return type or so?

j14159:
@lpil Indeed, `{}` is used to specify record types :)

I don't actually have a problem at all with expression function arg types `[in, a, list]` since list types themselves actually are expressed as `list 'the_contained_type` and, after all, Erlang does treat args as lists!

@OvermindDL1 `(int -> int) -> int` also seems pretty reasonable to me.  No danger of overloading  tuples and it does convey the...curry-ability...as well I think :)  I'm pretty easy between this and the list format @danabr has implemented so far but different directions with the auto-currying stuff might sway me more one way or the other.

PR #102, opened 2017-01-23
Bugfix: A FFI call must refer to existing variables.

danabr:


danabr:
Not entirely sure this belongs to the typer, but I placed it there as it is the typer that checks function arguments etc for the same thing.

j14159:
I think it's pretty reasonable to put it in the typer as I think it has a more complete view of the environment than the AST generation stage does.

Thanks for this!

PR #103, opened 2017-01-23
Bugfix: Missing code generation for the unit value.

danabr:


j14159:
Nice catch, thanks!

PS - I'm starting to get curious what  you're working on, finding these bugs :)

danabr:
I'm porting a small toy program to alpaca to see where the pain points are. I found FFI will actually be quite nice (even without any other constructs than `beam`) once we get top level function type declarations in.

Issue #62, opened 2016-12-07
Improve type and function sharing between modules


j14159:
Short list, driven by discussion on PR #61 with @danabr and from @lepoetemaudit's infix function work:

- all type _names_ available by default but not implementations (e.g. type constructors).  Exporting a type exposes its implementation.  This is to let us use a module's type in other modules' types without needing the details (information hiding, abstract/opaque types).
- individual functions used in other modules without qualifying them with their module name, this will make infix operators more useful.

Per @danabr we should consider a single `import` directive that allows importing specific functions, types, or even a subset of a type's constructors.

Ideas/expansions/criticisms most welcome.

j14159:
Opinions on separating function and type imports into two different statements?  E.g.

    import some_module.[foo/1, bar/3]

    import_type some_module.[baz, xyz]

Too much?  I kind of like the separation for clarity but I'm not particularly attached to it.

danabr:
Does not matter much to me.

Personally, I just want to bring a name into local scope, be it a function name or type name. That argues for having a single import statement.

However, since it is possible to define both a function `foo`, and a type `foo` in the same module, having two different statements is probably a good idea, since that removes unnecessary ambiguity.

yurrriq:
:+1: to `import` for functions and `import_type` for types.

yurrriq:
Related: http://docs.idris-lang.org/en/latest/tutorial/modules.html#meaning-for-data-types

j14159:
I kind of like the idea that a type itself can be completely hidden in a module as it seems with the Idris docs you linked, @yurrriq.

For type exports I'm thinking:
- without being explicitly exported, the type name isn't even visible to other modules.
- can export only the type name so that another module's type can leverage it without understanding its structure, equivalent to exposing an abstract type.  I think this addresses one of the things you were advocating @danabr but would appreciate feedback regardless.
- can export a subset of a type's constructors which also of course makes the type's name visible to other modules.
- some special form that exports the entire type and its members.

Type imports would be simpler, either just the type name or specific constructors/members.

The four export points do seem a little messy.  Might look something like:

    module either

    type t 'a = Left 'a | Right 'a

    -- only the type either.t is available, can't see Left/Right outside this module:
    export_type t

    -- either.t and either.Left are visible outside of this module:
    export_type Left

    -- either.t, either.Left, and either.Right are all visible externally:
    export_type t.*

[...]

PR #104, opened 2017-01-24
Fix 94

### LINKED ISSUE #94: Can't refer to a type qualified with a module name
Given

    module m
    type t = int

    module n
    type u = m.t

j14159:
Fixes #94 allowing us to qualify type constructors with their module name rather than always requiring a type to be imported.  This contributes to #62 by only allowing the use of constructors if their parent type has been exported by the module it's defined in.

### DIFF
--- a/src/alpaca_typer.erl
+++ b/src/alpaca_typer.erl
@@ -1640,6 +1640,25 @@ typ_of(Env, Lvl, #alpaca_apply{line=L, expr=Expr, args=Args}) ->
     %% (e.g. a symbol or bif), attempt to find it in the module.
     %% This ForwardFun function is used specifically to find functions defined
     %% later than the application we're trying to type.
+    CurryFun = 
+        fun(OriginalErr) ->
+            %% Attempt to find a curryable version
+            FN = case Expr of
+                    {symbol, Line, FunName} -> FunName;
+                    {bif, FunName, _, _, _} -> FunName
+            end,
+            CurryFuns = get_curryable_funs(Env#env.current_module, FN, length(Args)+1),
+            case CurryFuns of
+                [] -> OriginalErr;
+                [Item] -> case typ_of(Env, Lvl, Item) of
+                                {{t_arrow, TArgs, TRet}, NextVar} ->
+                                    {CurryArgs, RemArgs} = lists:split(length(Args), TArgs),
+                                    CurriedTypF = {t_arrow, CurryArgs, {t_arrow, RemArgs, TRet}},
+                                    typ_apply(Env, Lvl, CurriedTypF, NextVar, Args, L)
+                          end;
+                Items -> {error, {ambiguous_curry, Expr, Items, L}}
+            end
+    end,
     ForwardFun =
         fun() ->
                 FN = case Expr of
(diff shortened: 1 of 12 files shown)


src/alpaca_typer.erl
```diff
@@ -1009,9 +1013,49 @@ inst_type_arrow(EnvIn, {type_constructor, Line, Name}) ->
...
+    %% If the constructor was not qualified with a module name it's pretty
+    %% to fetch but if it was, then we need to first make sure it's in
+    %% the specified module *and* that its enclosing type is exported.
+    %%
+    %% Here's the easy local get:
+    GetTC = fun(#type_constructor{line=Line, name=Name, module=undefined}) ->
+                    Default = {error, {bad_constructor, Line, Name}},
+                    %% constructors defined in this module or imported by it:
+                    Available = EnvIn#env.type_constructors,
+                    C = proplists:get_value(Name, Available, Default);
```
danabr: Unused variable C (but maybe you added it for documentation purposes)?

src/alpaca_typer.erl
```diff
@@ -1009,9 +1013,49 @@ inst_type_arrow(EnvIn, {type_constructor, Line, Name}) ->
...
+                            %% similar manner.
+                            Types = Target#alpaca_module.types,
+                            F = fun(#alpaca_type{members=Ms}=AT) ->
+                                        [M#alpaca_constructor{type=AT} ||
+                                            #alpaca_constructor{
+                                               name=#type_constructor{name=TCN}
+                                              }=M <- Ms, TCN =:= Name]
+                                     end,
+                            case lists:flatten(lists:map(F, Types)) of
+                                [RealC|_] ->
```
danabr: Curious: When would you have multiple constructors with the same name?

j14159: Good point, the AST gen state does ensure unique constructors within a module.  I was being overly cautious I think :)

src/alpaca_typer.erl
j14159: No, definite mistake, thanks for catching that :)

src/alpaca_typer.erl
```diff
@@ -143,7 +143,7 @@ copy_cell(Cell, RefMap) ->
             {RevMembers, Map2} = lists:foldl(F, {[], RefMap}, Members),
             {RV2, Map3} = copy_cell(RV, Map2),
             {new_cell(#t_record{members=lists:reverse(RevMembers), row_var=RV2}), Map3};
-        #adt{vars=TypeVars, members=Members}=ADT ->
+        #adt{vars=TypeVars, members=Members, module=Mod}=ADT ->
```
j14159: I think there are a couple of warnings coming from leex and yecc that might be tricky to squash but I could be wrong.  We could certainly try it.

Regardless, thanks for pointing this out.

src/alpaca_typer.erl
```diff
@@ -986,13 +990,13 @@ inst_type_members(ADT, [_|T], Env, Memo) ->
...
                              {error, {bad_constructor_arg,term()}}.
-inst_type_arrow(EnvIn, {type_constructor, Line, Name}) ->
+inst_type_arrow(EnvIn, #type_constructor{}=TC) ->
     %% 20160603:  I have an awful lot of case ... of all over this
     %% codebase, trying a lineup of functions specific to this
     %% task here instead.  Sort of want Scala's `Try`.
     ADT_f = fun({error, _}=Err) ->
                     Err;
-               (#alpaca_constructor{type=#alpaca_type{}=T}=C) ->
+               (#alpaca_constructor{type=#alpaca_type{module=Mod}=T}=C) ->
```
j14159: Removing, thanks!

Issue #74, opened 2016-12-14
Automatic currying support


j14159:
Taken from the discussion in issue #56 with @danabr and @lepoetemaudit 

Single versions of a function will be automatically curried so the following will work:

    foo x y = x + y

    curry_foo () = 2 |> foo 1  -- results in 3

When there are different versions of the same named function (differing in arity), we will halt typing with an error in any ambiguous case.  For example the following would generate an error along the lines of {error, {ambiguous_application, foo/1, foo/2}} - but maybe less hostile than that :)

    foo x = x + x

    foo x y = x + y

    make_an_error () = 1 |> foo

While the following would not:

    foo x = x + x

    foo x y z = x + y + z

    -- unit -> (int -> int)
    passes_typing () = 2 |> foo 1

Feedback and differing opinions welcome.  I think basic expression application as discussed in #56 has to be addressed before this issue is.

lepoetemaudit:
I've been hacking in the typer and my initial impression is that it's going to be tricky to take the approach of getting the typer to understand curried applications. Not impossible, but there might be an easier way.

What if, when a given function has a single version defined, we auto-generated the other versions of the function? We could do that at the AST stage and then everything else should just 'work', i.e.

```elm
-- human defined a_fun/3
let a_fun x y z = (x, y, z)

-- autogenerated a_fun/1 and a_fun/2 variants
let a_fun x =
  let curried y z = a_fun x y z in curried

let a_fun x y =
  let curried z = a_fun x y z in curried
```

Obviously, we'd need to constrain it so that the curried functions aren't generated if you defined other variants. We could track which ones are actually used and strip unused ones at the codegen stage, or we could let these be 'opt-in' or 'opt-out' with some sort of sigil like

```elm
-- Curried
let ~my_fun x y z = (x, y, z)

-- Not curried
let my_other_fun x y z = (x, y, z)
```

(In traditional ML it's easier as you can always assume that an application involves exactly one argument, but as we have multiple arity we don't have that option)

j14159:
So if a module had only `f/3` we'd generate `f/2` and `f/1` but if it had `f/3` and `f/1` would we restrict it to those two or generate `f/2`?

Regardless the generation approach does solve the problem of the typer and codegen stages both needing to understand automatic currying which is pretty nice.

lepoetemaudit:
[...]

PR #105, opened 2017-01-24
Add currying

lepoetemaudit:
This PR enables currying, where there aren't conflicts between multi-arity functions and curried applications in the manner discussed in #74.

Currying is currently enabled only for top level functions, and far references must also be imported before being curried. In future these limitations should be easy enough to remove, and the addition of anonymous functions to Alpaca will also help see more benefit from currying.

The curries themselves are generated as anonymous functions at the call site to minimise polluting modules and conflicts.

If a user attempts to curry in a way that the compiler cannot resolve unambiguously (say, passing two arguments to function `f` when both `f/3` and `f/5` are defined) then the compiler provides an error, listing the ambiguous curries.

This PR also contains some alterations to the parser to allow infix arguments to take expressions as arguments, not simply terms. This means that in combination with currying infix functions can be used to delineate expressions, reducing parenthesis usage (e.g, chaining |>, or using <| as in Elm or $ as in Haskell to avoid needing to nest expressions in parentheses).

Here is an example (adapted from curry.alp in the test_files) that demonstrates some of what this now enables.

```elm
module curry

-- Infix 'pipe forward' operator
let (|>) x f = f x

let add x y = 
  x + y

let filter_ pred l acc =
  match l with
      [] -> acc
    | hd :: tail ->
        match pred hd with
            true  -> filter_ pred tail hd :: acc
          | false -> filter_ pred tail acc

let filter pred l =
  filter_ pred l []

-- Currying this means we can create an '=' predicate function for use in filter/2
let eq a b =
  a == b

let filtered_list () =    
  [1, 2, 3] -- without currying, the pipe below would fail
  |> filter (eq 2)

let foo () =
  let adder = add 6 in
  (6 |> add 5 |> add 5, adder 20, filtered_list ())
```

Shout outs to @danabr, @OvermindDL1, @yurrriq, @saem, and of course @j14159 for their considerable amount of thought and participation on this issue.

### DIFF
--- a/src/alpaca_ast.hrl
+++ b/src/alpaca_ast.hrl
@@ -134,7 +134,7 @@
 -type alpaca_binary() :: #alpaca_binary{}.
 
 -type alpaca_bits_type() :: int | float | binary | utf8.
-
+ 
 -record(alpaca_bits, {line=0 :: integer(),
                     %% Used to signal whether or not the bitstring is simply
                     %% using default size and unit values.  If it is *not*
@@ -379,6 +379,7 @@
 %%% Expressions that result in values:
 -type alpaca_value_expression() :: alpaca_const()
                                | alpaca_symbol()
+                               | alpaca_far_ref()
                                | alpaca_list()
                                | alpaca_binary()
                                | alpaca_map()
--- a/src/alpaca_ast.hrl
+++ b/src/alpaca_ast.hrl
@@ -109,9 +109,9 @@
 -type alpaca_bool() :: {bool, integer(), boolean()}.
 -type alpaca_atom() :: {atom, integer(), atom()}.
 
--type alpaca_error() :: {raise_error, 
-                       integer(), 
-                       throw|error|exit, 
+-type alpaca_error() :: {raise_error,
+                       integer(),
+                       throw|error|exit,
                        alpaca_value_expression()}.
 
 %%% The variable _, meaning "don't care":
@@ -134,7 +134,7 @@
 -type alpaca_binary() :: #alpaca_binary{}.
 
 -type alpaca_bits_type() :: int | float | binary | utf8.
- 
+
 -record(alpaca_bits, {line=0 :: integer(),
                     %% Used to signal whether or not the bitstring is simply
                     %% using default size and unit values.  If it is *not*
@@ -197,7 +197,7 @@
 
 -record(t_record, {members=[] :: list(t_record_member()),
                            row_var=undefined :: typ()}).
-                           
+
 -type t_record() :: #t_record{}.
 
 %%% ADT Type Tracking
--- a/src/alpaca_codegen.erl
+++ b/src/alpaca_codegen.erl
@@ -346,7 +346,7 @@ gen_expr(Env, #alpaca_apply{line=L, expr=Expr, args=Args}) ->
     Env2 = Env#env{synthetic_fun_num=Env#env.synthetic_fun_num + 1},
     case Expr of
         %% Detect far refs that require currying
-        {alpaca_far_ref, _, _, _, Arity} when Arity > length(Args) ->
+        #alpaca_far_ref{arity=Arity} when Arity > length(Args) ->
             CArgs = lists:map(
                fun(A) -> 
                     {symbol, L, "carg_" ++ integer_to_list(A)}
--- a/src/alpaca_codegen.erl
+++ b/src/alpaca_codegen.erl
@@ -742,4 +742,17 @@ module_info_helpers_test() ->
     ?assert(is_list(Mod:module_info())),
     true = code:delete(Mod).
 
+curry_test() ->
+    Code = 
+        "module curry\n"
+        "let f x y = x + y\n"
+        "let main () = \n"
+        "  let f_ = f 10 in\n"
+        "  f_ 6",
+    {ok, _, Bin} = parse_and_gen(Code),
+    Mod = curry,
+    {module, Mod} = code:load_binary(Mod, "alpaca_curry.beam", Bin),
+    ?assertEqual(Mod:main(unit), 16),
+    true = code:delete(Mod).
+
 -endif.
--- a/src/alpaca_codegen.erl
+++ b/src/alpaca_codegen.erl
@@ -28,11 +28,11 @@
 %%   - names of top-level functions with their arity
 %%   - incrementing variable number for wildcard variables (underscores)
 %%   - numbers for synthesized function name generation
-%% 
-%% The top-level functions get looked up for correct Core Erlang call 
+%%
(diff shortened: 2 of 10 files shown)


j14159:
@lepoetemaudit could you please include a section in either the `README.md` or `Tour.org` files to talk a little about how it works from the user perspective?

Otherwise I love this :)

src/alpaca_ast.hrl
```diff
@@ -134,7 +134,7 @@
 -type alpaca_binary() :: #alpaca_binary{}.
 
 -type alpaca_bits_type() :: int | float | binary | utf8.
-
+
```
danabr: Nitpick: Trailing whitespace.

src/alpaca_codegen.erl
```diff
@@ -113,7 +113,7 @@ gen_fun(Env, #alpaca_fun_def{name={symbol, _, N}, versions=Vs}=Def) ->
                 false ->
                     FName = cerl:c_fname(list_to_atom(N), length(Args)),
                     A = [cerl:c_var(list_to_atom(X)) || {symbol, _, X} <- Args],
-                    {_, B} = gen_expr(Env, Body),
+                    {_, B} = gen_expr(Env, Body),
```
danabr: Nitpick: Trailing whitespace.

src/alpaca_codegen.erl
```diff
@@ -290,17 +291,52 @@ gen_expr(Env, #alpaca_apply{expr={symbol, _Line, Name}, args=[{unit, _}]}) ->
...
+    {FName, Curry, Arity} = case proplists:get_all_values(Name, Env#env.module_funs) of
+        [] -> {cerl:c_var(list_to_atom(Name)), false, 0};
+        AvailFuns ->
+            %% If we have an exact arity match, use that, otherwise curry
+            case lists:filter(fun(X) -> X =:= DesiredArity end, AvailFuns) of
+                [A] -> {cerl:c_fname(list_to_atom(Name), A), false, A};
+                _ ->  
+                    PossibleCurries = lists:filter(fun(X) -> X > DesiredArity end, AvailFuns),
+                    %% The typer ensures that we can curry unambiguously
+                    case PossibleCurries of
```
danabr: Not strictly necessary with a case here, as as you say the typer already guarantees it will be an unambiguous match. Hence: 
````
%% The typer ensures that we can curry unambiguously
[CurryArity] = lists:filter(fun(X) -> X > DesiredArity end, AvailFuns),
{cerl:c_fname(list_to_atom(Name), CurryArity), true, CurryArity}
````

File: src/alpaca_codegen.erl [hunk 4]
```diff
@@ -290,17 +291,52 @@ gen_expr(Env, #alpaca_apply{expr={symbol, _Line, Name}, args=[{unit, _}]}) ->
...
+           Binding = #fun_binding{
+                        expr={symbol, L, CurryFunName},
+                        def=CurryExpr},
+
+           gen_expr(Env2, Binding);
+
+        false -> Apply = cerl:c_apply(
+                    FName, 
+                    [A || {_, A} <- [gen_expr(Env, E) || E <- Args]]),
+    {               Env, Apply}
```
danabr: Not entirely sure what happened with the formatting here?

src/alpaca_codegen.erl
```diff
@@ -290,17 +291,52 @@ gen_expr(Env, #alpaca_apply{expr={symbol, _Line, Name}, args=[{unit, _}]}) ->
...
+                                            line=L,
+                                            expr=FExpr,
+                                            args=Args ++ CArgs}}]},
+           Binding = #fun_binding{
+                        expr={symbol, L, CurryFunName},
+                        def=CurryExpr},
+
+           gen_expr(Env2, Binding);
+
+        false -> Apply = cerl:c_apply(
```
danabr: This would probably be more readable if you pushed "Apply =" down to the next line.

src/alpaca_typer.erl
```diff
@@ -1606,28 +1640,37 @@ typ_of(Env, Lvl, #alpaca_apply{line=L, expr=Expr, args=Args}) ->
...
                             {error, _}=Err -> Err;
                             {TypF, NextVar} ->
-                                typ_apply(Env, Lvl, TypF, NextVar, Args, L)
+                                %% We should have a t_arrow taking some args with a return value
+                                %% What we need is a t_arrow that takes some of those args and returns
+                                %% another t_arrow taking the remainder and returning the final arg
+                                try
+                                    typ_apply(Env, Lvl, TypF, NextVar, Args, L)
+                                catch
+                                    error:{arity_error, _, _} = Err -> CurryFun(Err)
```
danabr: Since `{arity_error, _, _}` is module internal and used for control flow, it is more idiomatic to use `throw` rather than `error`.

src/alpaca_typer.erl
```diff
@@ -1606,28 +1640,37 @@ typ_of(Env, Lvl, #alpaca_apply{line=L, expr=Expr, args=Args}) ->
...
                         end;
-                    {error, _} = E -> E
+                    {error, _} = E -> CurryFun(E)
                 end
         end,
 
     case typ_of(Env, Lvl, Expr) of
         {error, {bad_variable_name, _}} -> ForwardFun();
-        {error, _} = E -> E;
+        {error, _} = E -> E;
```
danabr: Nitpick: Trailing whitespace.

lepoetemaudit:
@danabr - thanks for spotting all these, I will amend

j14159:
@lepoetemaudit let me know when you're happy with this and I'll merge it and rebase mine again.

lepoetemaudit:
@j14159 sure, I just need to finish amending from @danabr's feedback and umm, figure out enough Emacs to export to markdown from `tour.org` :)

j14159:
`M-x org-md-export-to-markdown` should do the trick if memory serves :)

lepoetemaudit:
That works, alas it gives me anchors like `<a id="orge61d414"></a>` instead of the nice ones like `<a href="#sec-1">`. Tried updating everything I could to no avail. I imagine this is down to my ignorance (or maybe it just knows I'm a Vim user and it's trolling me...)

j14159:
Hrm, ok.  I don't think I modified the output of it when I last exported it.  Feel free to skip updating the markdown and I can do it in a separate PR after I rebase if you like.  Not a big deal.

lepoetemaudit:
I just manually synced the markdown file as it was only a small addition (and spotted a mistake I added in the tour as I did it), and I've fixed everything else. I enabled whitespace stripping in my editor and it's added a bit of noise to the diffs, so apologies for that.

j14159:
Thanks again for this!

PR #107, opened 2017-01-25
Retrying PR with rebase

### LINKED ISSUE #94: Can't refer to a type qualified with a module name

j14159:
FIxes #94

Had some conflicts after rebasing on @danabr and @lepoetemaudit's work so retrying the PR from a clean state.

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,6 +328,7 @@ modules in Erlang with top-level entities including:
 - type declarations (ADTs)
 - functions which can contain other functions and variables via `let`
 bindings.
+- functions are automatically curried (with some limitations)
 - simple test definitions
 
 An example:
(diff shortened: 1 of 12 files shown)


danabr:
The commit history looks strange. Did you really do a successful rebase? You should have had way less commits then.

j14159:
There were a number of conflicts since a few of us were in parser and typer at what git thinks is the same time it appears but I did rebase.  Agreed it looks a little messy...suggestions?

danabr:
At this point I would:
git checkout origin/master
git checkout -b fix-92-squashed
git merge --squash fix-92
git commit
git push origin HEAD

That is, squash it all into one commit. You lose some history, but I think the changes are small enough for it to work.

j14159:
That's pretty reasonable.  I think I know what I might've messed up as well so I may try to rewind this somewhere else first.

PR #108, opened 2017-01-26
Reduce the number of reserved words related to binaries.

danabr:
This removes the following words as reserved words in the language:
* unit
* size
* end
* sign
* big
* little
* native
* utf8

"unit" in particular is needed to refer to the unit type. That change is
not included in this changeset, though.


j14159:
Good call, thanks for this!

PR #109, opened 2017-01-26
Bugfix: Added the ability to refer to 'chars' and '()'  in type specs.

danabr:
Also made sure these are accepted as literals.

### DIFF
--- a/test_files/values.alp
+++ b/test_files/values.alp
@@ -2,9 +2,12 @@ module values
 
 export test_values/1
 
-type simple_adt = AnInt int | AString string
+type simple_adt = AnInt int | AString string | AFloat float |
+                  ARecord { x: int } | ABinary binary | AList (list int) |
+                  AnADT simple_adt | AnAtom atom | ABool bool | Chars chars |
+                  ATuple (int, string) | Unit ()
 
-let test_int = 41 
+let test_int = 41
 let test_string = "Vicugna pacos"
 let test_float = 0.5
 let test_record = {x = 19}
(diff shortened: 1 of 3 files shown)


test_files/values.alp
```diff
@@ -15,6 +18,20 @@ let test_atom = :nope
 let test_bool = true
 let test_chars = c"x"
 let test_tuple = (100, "100")
+let test_unit = ()
+
+let wrapped_int = AnInt 41
```
danabr: @lepoetemaudit: I tried writing
````
````
but then I got `{error,non_literal_value, ...`.  Perhaps we should allow arbitrary names in here, and resolve the references in alpaca_ast_gen. What do you think?

j14159: @danabr @lepoetemaudit I think maybe we just need to allow for constructors as literal values?  The typer should catch any further issues.

danabr: Or rather allow symbols as literal values (literal is not as fitting a name then, perhaps), and then have the typer or ast_gen stage complain if the symbol doesn't refer to a zero arity function.

Constructors as literals are already supported. as the test witnesses.

j14159: Quite right, apologies!  I would blame insufficient coffee this morning but really I should have just looked closer :)

@danabr if you want this merged prior to discussion/changes between you and @lepoetemaudit  or prefer I hold off just let me know.  I'm fine either way.

danabr: I'm fine with merging this. The purpose of this PR was to add `chars`and `()` as basic types. I just happened to notice the described behavior when implementing this and wondered if it was intentional.

j14159: OK, will merge for now.  Thanks!

PR #111, opened 2017-01-27
Module qualified type constructors

### LINKED ISSUE #94: Can't refer to a type qualified with a module name

j14159:
Fixes #94 

cc/ @danabr for review, branched off master and cherry-picked commits as there were a small number.

### DIFF
--- a/src/alpaca_parser.yrl
+++ b/src/alpaca_parser.yrl
@@ -20,7 +20,7 @@ const module_fun
 type poly_type poly_type_decl type_vars type_member type_members type_expr
 sub_type_expr type_expressions type_tuple comma_separated_type_list type_list
 module_qualified_type module_qualified_type_name
-type_apply
+type_apply module_qualified_type_constructor
 type_import type_export types_to_export
 
 test_case
@@ -246,6 +246,16 @@ poly_type_decl -> symbol type_vars :
 
 type_vars -> type_var : ['$1'].
 type_vars -> type_var type_vars : ['$1'|'$2'].
+
+module_qualified_type_constructor -> symbol '.' type_constructor :
+  {symbol, _, Mod} = '$1',
+  {type_constructor, L, N} = '$3',
+  #type_constructor{line=L, module=list_to_atom(Mod), name=N}.
+
+type_apply -> module_qualified_type_constructor term :
+  #alpaca_type_apply{name='$1', arg='$2'}.
+type_apply -> module_qualified_type_constructor :
+  #alpaca_type_apply{name='$1'}.
 
 type_apply -> type_constructor term :
   {type_constructor, L, N} = '$1',
--- a/src/alpaca_parser.yrl
+++ b/src/alpaca_parser.yrl
@@ -19,6 +19,7 @@ op infix
 const module_fun
 type poly_type poly_type_decl type_vars type_member type_members type_expr
 sub_type_expr type_expressions type_tuple comma_separated_type_list type_list
+module_qualified_type module_qualified_type_name
 type_apply
 type_import type_export types_to_export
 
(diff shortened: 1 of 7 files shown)


src/alpaca_parser.yrl
```diff
@@ -145,12 +150,36 @@ record_type_members -> record_type_member ',' record_type_members : ['$1' | '$3'
...
+     module = list_to_atom(Mod),
+     name = {type_name, L, Name}}.
+
+module_qualified_type -> module_qualified_type_name type_expressions:
+  {module_qualified_type_name, L, Mod, Name} = '$1',
+  #alpaca_type{
+     line = L,
+     module = list_to_atom(Mod),
+     name = {type_name, L, Name},
+     vars = [V || {type_var, _, _}=V <- '$2']}.
```
danabr: Don't you need to set `members` as well. E.g. `type x = option.option int` would yield `#alpaca_type {members=[t_int]}`, right?

j14159: `int` would not necessarily be a member here though, it would be the value assigned to a type variable in `option.option` and we don't actually have any information about what `option.option`'s full membership is at this point - I think the typer needs to instantiate those.  E.g. the following works without requiring the members to be added in the parser (and I'll add this to a test):

    module m
    export_type t
    type t 'a = T 'a

    module n
    type u 'a = U m.t 'a
    let f () = U m.T 1

But in looking at this I did find another bug!  This does _not_ work, it tries to unify `t_int` with `undefined`:

    module n
    type u = U m.t int
    let f () = U m.T 1

And interestingly this exhibits the exact same unification failure:

    module n
    type opt 'a = Some 'a | None
    type u = U opt int
    let f () = U Some 1

Adding the members in the parser stage does not seem to correct this in the latter example and it seems to me that neither of those should fail.  I haven't dug into this yet but I expect what's happening is that when we do:

    type option 'a = Some 'a | None
    type something_else = option int

The parser has no idea that `int` is supposed to be assigned to a variable `'a` in that second line and so trying to get that variable from the proplist in an `#adt{}` yields `undefined`.  That's a big guess though as I haven't looked too closely.  I'll file this as a new issue.

I suspect this might be as simple as the parser manufacturing variable names for any type parameters that are not explicitly type variables.

danabr: But don't we lose information if we don't register that "type int_option" is the type option.option instantiated with an int argument? Not just any value of any type would do, it should be specifically ints. Right now it looks like we drop that information on the floor. What am I missing?

danabr: Your example works because you instantiate the type with a type variable, and those you *do* record on the line above. I meant to highlight that we need to support instantiating with concrete types as well. But I think you maybe addressed that with #113.

j14159: You're not missing anything, you're absolutely correct that we're dropping that information but it belongs in a type variable rather than as a member I think.  E.g.

    type opt 'a = Some 'a | None

Here `'a` is not actually a member of the type, it's the argument to `Some` which along with `None` are the members.  Whereas in the following:

    type something_else 'a = 'a | What 'a

`'a` is indeed a member.  I filed issue #113 for this but I'm entirely willing to believe it's not describing this sufficiently :)  Thoughts?

j14159: @danabr haha, we're overlapping here.  You're right, I think that #113 addresses this and I completely agree with you that instantiating with concrete types is necessary.

danabr: Yes, I know it is not technically a member. But at least in some cases I've encountered type variables get registered as members, and then matched somewhere in the typer. I think it is in `inst_constructor_arg`, and specifically the call to `replace_vars`. I thought the same process should work here as well.

j14159:
Ended up with a merge in there because of some messy disagreements with git between my local master rebased against upstream and my origin/master.

PR #112, opened 2017-01-28
Fix ambigous parsing of higher order types.

### LINKED ISSUE #110: Type expression parsed differently whether a builtin type is used or not.
This code parses correctly:
````
type my_map = map atom atom
````

This one fails with `["syntax error before: ",[]]`:
````
type my_atom = atom
type my_map = map my_atom atom
````

This again parses fine:
````
type my_atom = atom
type my_map = map (my_atom) atom
````

i.e, `type my_map = map my_atom atom` is parsed as `type my_map = map (my_atom atom)` .

danabr:
`type x = foo bar baz` is now parsed as just `type x = (foo bar baz)`, not as  `type x = foo (bar (baz))`.

Closes #110.

### DIFF
--- a/src/alpaca_ast_gen.erl
+++ b/src/alpaca_ast_gen.erl
@@ -1594,6 +1594,48 @@ type_expr_in_type_declaration_test() ->
     ?assertMatch({error, _}, test_parse("type a not_a_var = A not_a_var")).
 
 
+ambiguous_type_expressions_test() ->
+    ?assertMatch({ok, {alpaca_type,undefined,
+                         {type_name,1,"my_map"},
+                         [],
+                         [{alpaca_map,
+                              {alpaca_type,undefined,
+                                  {type_name,1,"foo"},
+                                  [],[]},
+                              t_atom}]}},
+                 test_parse("type my_map = map foo atom")),
+    ?assertMatch({error, _}, test_parse("type my_map = map foo bar atom")),
+    ?assertMatch({error, _}, test_parse("type my_list = list foo atom")),
+    ?assertMatch({error, _}, test_parse("type my_pid = pid foo atom")),
+    ?assertMatch({ok, {alpaca_type,undefined,
+                         {type_name,1,"my_type"},
+                         [],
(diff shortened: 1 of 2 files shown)


j14159:
I'm not sure why it's not showing here but it looks like we have a test failure with CT:  https://travis-ci.org/alpaca-lang/alpaca/builds/196143693

danabr:
It specifies the wrong commit hash though (037a90e...). The correct one is ff005f2fa2bfa5e76bcf0612e7e264ad5150354b. All tests pass locally for me.

j14159:
Right, sorry about that.  I wonder why the connection to travis is broken here...

danabr:
Yes, it is a bit weird. Maybe because I rebased. I normally don't do that once I've opened a pull request (mean to the reviewer), but I wanted to fix the commit message.

j14159:
No worries, will merge now.  Thanks for this!

danabr:
Argh, there *was* a real test failure. I should have looked at the actual log output. The failure is in the common test suite, which I don't normally run locally. I'll start doing that and  I'll see if I can fix whatever broke.

danabr:
The test actually generates an invalid type specification;
````
failed to compile:
<<"module a ;; type c = pid list atom | A ;; type a = B ;;">>
````
pid takes a single argument. To be consistent with how other polymorphic types are interpreted, this should yield a parse error. i.e. if you switch out `pid` for `foo` in the example, it would be parsed as a type foo taking two arguments. I wanted the bultin types to be consistent with this. I'll see if I can fix the generator on Monday. I have to upgrade my OTP installation to 19 first to run the tests and then I need a proper internet connection.

j14159:
@danabr master is failing during `rebar3 ct`:  https://travis-ci.org/alpaca-lang/alpaca/builds/196227950

I think the proper tests need to be updated to properly generate things like `foo list atom` now as `foo (list atom)`.

j14159:
No problem, maybe @ypaq has some ideas to fix this easily as well.

PR #114, opened 2017-01-28
Fix property tests generating ambigous type specs.

danabr:
While at it, add the `chars`and `()` types as base types.


danabr:
And again, sorry for messing up the build.

PR #119, opened 2017-02-01
Remove {alpaca_list, T} in favor of {t_list, T}.

danabr:
Having two representations is confusing.


danabr:
@j14159: Maybe you had some good reason for doing like this. If so, I will have to patch the exhaustiveness checker instead, because it only checks for the {t_list, T} case.

j14159:
No good reason for me doing it that I can recall, good change :)  Thanks for this!

PR #120, opened 2017-02-01
Bugfix 91

### LINKED ISSUE #91: Referring to missing types does not generate an error
module x

    type a = int | b

This doesn't generate an error about an undefined type `b` and it should.

Thanks to @danabr for pointing this out in IRC

j14159:
Fixes #91 

This should catch type members that are:

- bad variable names
- references to types that don't exist in the module or imports
- references to module-qualified types that don't exist

### DIFF
--- a/src/alpaca_typer.erl
+++ b/src/alpaca_typer.erl
@@ -1431,7 +1431,7 @@ validate_types(
     #alpaca_type{name={_, L, N}, vars=Vs, members=Ms}=H,
     case [TN || TN <- TypeNames, TN =:= N] of
         [] ->
-            throw({invalid_type, ModName, L, N});
+            throw({unknown_type, ModName, L, N});
         _ ->
             validate_types(ModName, TypeNames, Modules, Ms),
             validate_types(ModName, TypeNames, Modules, Vs),
@@ -1445,7 +1445,7 @@ validate_types(MN, Ts, Mods, [#alpaca_type{}=T|Rem]) ->
         [#alpaca_module{types=Xs}] ->
             case [X || #alpaca_type{name={_, _, Y}}=X <- Xs, Y =:= N] of
                 [] -> 
-                    throw({invalid_type, MN, L, N});
+                    throw({unknown_type, MN, L, N});
                 [_] ->
                     [] = validate_types(MN, Ts, Mods, Vs),
                     validate_types(MN, Ts, Mods, Rem)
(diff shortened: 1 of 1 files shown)


src/alpaca_typer.erl
```diff
@@ -1404,6 +1413,67 @@ type_module_tests([], _, _, Funs) ->
...
+ validate_types(ModName, TypeNames, Mods, [#alpaca_type_tuple{members=Ms}|T]) ->
+    validate_types(ModName, TypeNames, Mods, Ms),
+    validate_types(ModName, TypeNames, Mods, T);
+validate_types(MN, Ts, Mods, [#t_record{members=Ms}|T]) ->
+    MemberTypes = [Type || #t_record_member{type=Type} <- Ms],
+    validate_types(MN, Ts, Mods, MemberTypes ++ T);
+validate_types(M, TNs, Mods, [{t_list, LT}|T]) ->
+    [] = validate_types(M, TNs, Mods, [LT|T]),
+    validate_types(M, TNs, Mods, T);
+validate_types(M, TNs, Mods, [{alpaca_map, KT, VT}|T]) ->
```
danabr: Here is another example of a duplicate type representation: `{alpaca_map, K, V}` vs `{t_map, K, V}`.

src/alpaca_typer.erl
```diff
@@ -1404,6 +1413,67 @@ type_module_tests([], _, _, Funs) ->
...
+    end;
+validate_types(MN, Ts, Mods, [#alpaca_type{}=T|Rem]) ->
+    #alpaca_type{name={_, L, N}, module=TargetMod, vars=Vs} = T,
+    case [M || #alpaca_module{name=X}=M <- Mods, X =:= TargetMod] of
+        [] -> 
+            throw({bad_module, MN, L, TargetMod});
+        [#alpaca_module{types=Xs}] ->
+            case [X || #alpaca_type{name={_, _, Y}}=X <- Xs, Y =:= N] of
+                [] -> 
+                    throw({invalid_type, MN, L, N});
```
danabr: Maybe `unknown_type` rather than `invalid_type`?

src/alpaca_typer.erl
j14159: Good point, mind filing a quick issue for that?

src/alpaca_typer.erl
j14159: Works for me, I'll make that change and merge.  Thanks!

src/alpaca_typer.erl
danabr: Submitted #121

j14159: Thanks!

PR #122, opened 2017-02-02
Unify error handling in alpaca_ast_gen.

danabr:
* All errors in `alpaca_ast_gen` are returned as `{error, Reason}` tuples.
* All possible errors are documented in type specs.
* Errors are propagated to the caller of `alpaca:compile/2` via return value.

### DIFF
--- a/src/alpaca.erl
+++ b/src/alpaca.erl
@@ -57,24 +57,30 @@ compile(What) ->
     compile(What, []).
 
 compile({text, Code}, Opts) ->
-    Mods = alpaca_ast_gen:make_modules([Code]),
-    case type_modules(Mods) of
-        {error, _}=Err -> Err;
-        {ok, [TypedMod]} ->
-            {ok, Forms} = alpaca_codegen:gen(TypedMod, Opts),
-            compile:forms(Forms, [report, verbose, from_core])
-    end;
+    compile_phase_1([{"<no file>", Code}], Opts);
 
 compile({files, Filenames}, Opts) ->
-    Code = load_files(Filenames),
-    {ok, Mods} = type_modules(alpaca_ast_gen:make_modules(Code)),
+    compile_phase_1(load_files(Filenames), Opts).
+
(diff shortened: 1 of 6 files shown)


danabr:
When working on this, I noticed a number of areas for improvement:
* For some errors we report line numbers, for other we don't. We should consistently report line numbers when available.
* It would be useful to store the file name among with `#alpaca_module{}`. This way, we could always refer to the relevant filename in the errors.

Also note that I only updated the `alpaca` module to propagate the errors from `alpaca_ast_gen`. It should be further refactored as error reporting is consolidated for the other phases.

j14159:
This is great, thanks for this!  I love seeing all the code deletes! :)  I agree with both of your points as well.

Minor concern with the different compiler phases, I think we could wrap the stages as named functions in a list of passes with some wrapper function that handles short-circuiting on errors - this may already be what you're thinking of with the compiler framework and I've seen this in the LFE source as well, e.g.

    compile({files, Fs}, Opts) ->
        Phases = [fun ast_gen/2, fun typer/2, fun exhaustive_matches/2, fun binaries_out/2],
        run_phases(Phases, Opts, none).

    run_phases([], _Opts, LastResult) ->
        LastResult;
    run_phases([Next|Rem], Opts, LastResult) ->
        run_phases(Rem, Opts, phase(Next, Opts, LastResult)).

    phase({error, _}=Err, _, _) ->
        Err;
    phase({ok, PastSuccess}, Opts, NextStage) ->
        NextStage(PastSuccess, Opts).

There are probably a few issues with the above, just what came to mind between the LFE sources and what you were suggesting on the mailing list before.  Thoughts?  I'm quite happy to merge this as-is btw.

danabr:
Tried to reply by email, but that apparently did not work...

What you suggest is pretty much what I had in mind as well. My thinking was to convert the rest of the phases to follow the same style before doing that, though.

j14159:
Works for me!  Merging.  Thanks again :)

PR #124, opened 2017-02-06
Functional record updates

### LINKED ISSUE #45: Updates to records
Need a way to update or add fields to records, e.g.

    let r = {x=1, y=2} in
      {r | z=3}

j14159:
Fixes #45 

Pretty basic syntax to add or update a field, e.g to add an `x: int` field to an existing record:

    {x=1 | existing_record}

Since record updates are non-destructive and thus not (strictly speaking) updates, I've allowed for members' types to change, e.g. the following is legal:

    -- results in {x: int, y: int}
    {x=1 | {x=1.0, y=5}}

I think this makes sense both because our updates are non-destructive and since we allow for records to be declared anywhere as literals without them specified as types prior.  This PR includes a test showing information loss when using type constructors for which I'll file a new issue.  I'm not entirely certain it's something we want to consider a real problem but I'll leave it for discussion in the issue itself once filed.

### DIFF
--- a/src/alpaca.erl
+++ b/src/alpaca.erl
@@ -411,4 +411,14 @@ throw_with_variables_test() ->
     ?assertThrow({not_equal, 1, 2}, M:assert_equal(1, 2)),
     code:delete(M).
 
+record_update_test() ->
+    Files = ["test_files/update_record.alp",
+             "test_files/use_update_record.alp"],
+    [M1, M2] = compile_and_load(Files, []),
+
+    ?assertEqual(#{'__struct__' => record, x => 5}, M2:main(unit)),
+
+    code:delete(M1),
+    code:delete(M2).
+
 -endif.
--- a/src/alpaca.erl
+++ b/src/alpaca.erl
@@ -169,7 +169,7 @@ type_import_test() ->
     ModuleNames = compile_and_load(Files, []),
     io:format("Compiled and loaded modules are ~w~n", [ModuleNames]),
     M = alpaca_type_import,
-    ?assertEqual(2, M:test_output(unit)),
+    ?assertEqual(2, M:test_output({})),
     [code:delete(N) || N <- ModuleNames].
 
 type_imports_and_pattern_test() ->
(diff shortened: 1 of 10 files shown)


j14159:
I opened issue #125 for the information loss problem.

src/alpaca.erl
```diff
@@ -411,4 +411,24 @@ throw_with_variables_test() ->
     ?assertThrow({not_equal, 1, 2}, M:assert_equal(1, 2)),
     code:delete(M).
 
+record_update_test() ->
+    Files = ["test_files/update_record.alp",
+             "test_files/use_update_record.alp"],
+    [M1, M2] = compile_and_load(Files, []),
+
+    ?assertEqual(#{'__struct__' => record, x => 5, b => 2}, M2:main(unit)),
```
danabr: I find it a bit confusing to use `unit` as the parameter in a call to an alpaca function `let main () = ...`, when the actual representation of unit is the empty tuple `()` (or `{}` in Erlang syntax). Speaking of this, shouldn't "let main () = " be compiled to a pattern match against `{}`, so this would crash?

src/alpaca_typer.erl
```diff
@@ -1037,6 +1068,8 @@ inst_type_arrow(EnvIn, #type_constructor{}=TC) ->
                     Default = {error, {bad_constructor, Line, Name}},
                     %% constructors defined in this module or imported by it:
                     Available = EnvIn#env.type_constructors,
+                    io:format("Get constructor for ~s~n", [Name]),
```
danabr: Do you want to keep this debug output?

danabr:
I wouldn't call it an "update" if it changes the type of the record. A better name would be record transformation. The mental model then (using your example) is that it is a shortcut for writing a function `{x: int, y:int} -> {x:float, y:int}`.

src/alpaca_typer.erl
j14159: Definitely not, thanks for catching that :)

j14159:
I like "record transformation", thanks for the suggestion.  I'll update docs and AST nodes (e.g. `alpaca_record_update` -> `alpaca_record_transform`).

src/alpaca.erl
j14159: We could do that.  I've always just treated it as a throwaway value and not bothered to check.  I don't have particularly strong opinions on this but our code gen _does_ output an empty Erlang tuple.  I suppose this is more of an issue for Erlang calling Alpaca.

I'll change this particular occurrence to an empty tuple, happy to have the broader "do we crash on anything not an empty tuple" in a separate issue.

PR #126, opened 2017-02-06
Make `map` not a reserved word

j14159:
This should be left open for data structures that are basically
functors, e.g. option, etc.


danabr:
Nice! I've been planning to do the same for all builtin types for a while.

j14159:
Probably a really good idea.  I just wanted to make sure `map` was available for use in https://github.com/alpaca-lang/alpaca_lib right away ;)

PR #129, opened 2017-02-07
Fixes arity of generated curries

lepoetemaudit:
In some cases, codegen would fail as the arity of the generated function was not correctly calculated. For example, `let f x y z = x + y + z;; let main () = let curry = f 10 in curry 20 30` would run into this.

Sorry for not spotting this before, looks my tests weren't exhaustive enough.


j14159:
No worries, it's still pretty early days :)  Thanks for catching this!

PR #130, opened 2017-02-07
Remove base types as reserved words.

danabr:
This frees up the following to be used as identifiers in code:
* atom
* binary
* bool
* chars
* float
* list
* pid
* string


j14159:
This is great, thanks!

PR #132, opened 2017-02-08
Provide file path and line number for all parse errors.

danabr:
This makes all errors from the parse and AST generation stages contain the file name/path and the line number the error pertains to.

There are two cases were line numbers are incorrectly reported (i.e. reported as 1) at the moment:
* `{invalid_top_level_construct, term()}`. We can hopefully remove this in the future by restructuring the grammar. Another alternative is to *consistently* specify the line numbers for all elements generated by the parser. Both of these refactorings are too big to include in this work.
* `{no_module, module()}` when an unknown module is specified in a type import. I tried this, but the required changes were a bit too invasive to be included at this time.

### DIFF
--- a/src/alpaca_ast_gen.erl
+++ b/src/alpaca_ast_gen.erl
@@ -22,15 +22,15 @@
 
 -type error_reason() :: {duplicate_definition, string()} |
                         {duplicate_type, string()} |
+                        {function_not_exported, module(), string()} |
                         {invalid_endianess, term()} |
                         {invalid_bin_qualifier, string()} |
                         {invalid_bin_type, string()} |
                         {invalid_fun_parameter, term()} |
-                        {module_not_found, module()} |
                         {module_rename, module(), module()} |
                         no_module |
+                        {no_module, module()} |
                         {syntax_error, string()} |
-                        {type_not_exported, module(), string()} |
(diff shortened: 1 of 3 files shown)


danabr:
After this, I plan to add an "alpaca_error_fmt" module, that provides a nice error messages for all of these errors in a given locale. Then it is time to make all the other phases follow the same error handling pattern as "alpaca_ast_gen" now does.

src/alpaca_ast_gen.erl
```diff
@@ -1662,31 +1674,31 @@ invalid_map_type_parameters_test() ->
             "type x = map int",
     Code3 = "module a\n\n"
             "type x = map int int int",
-    ?assertMatch({error,{parse_error,_, {_,{wrong_type_arity,t_map, 0}}}},
+    ?assertMatch({error,{parse_error, ?FILE, 3, {wrong_type_arity,t_map, 0}}},
```
j14159: Out of curiosity where is `?FILE` coming from?

j14159:
Works for me, thanks for this!

src/alpaca_ast_gen.erl
danabr: See `test_make_modules/1`. `?FILE` is just a builtin erlang macro that gives you the filename of the current file. I just use that as the source of the modules that I define in the tests.

Issue #95, opened 2017-01-17
Guards in function heads


danabr:
Trying to compile @j14159's example code from another issue:
````
module guards
type make_it_work = int | string
let f x, is_int x = x + 1
let f x, is_string s = string_append "hello, x"
````
we get `{error,{3,alpaca_parser,["syntax error before: ","','"]}}`

For consistency with `match` expresssions, guards should be allowed in function heads (just like in Erlang).

Side note: Personally, I would rather drop guards completely from `match` expressions, since guards disable exhaustiveness checks ("Having a compiler warn about non-exaustive guards would be impossible in the general case, as it would involve solving the halting problem" (http://stackoverflow.com/a/7109455/347687)), and rather have an if expression.

j14159:
Interesting link, thanks for posting that!  While I follow the idea of things like `==`, `<`, etc in terms of inhibiting proper exhaustiveness checks it's not clear to me how things like the type check guards would do the same.  Wouldn't these actually assist exhaustiveness checks with sum/union types?

I'm not at all opposed to introducing an `if ... then ... else` construct but I do wonder if we should maybe allow the full range of guards and throw an additional warning on the ones that are _not_ type checks, saying something like "you just disabled exhaustiveness checks because ==".  I'm basically worried about reducing expressiveness, see issue #86.  Having said that, I recognize that I may be paradoxically arguing for a reduction in safety here.  Perhaps it would make sense to allow the type checks for dispatch stuff but disallow the rest?  What do you think?

danabr:
As you say, the type check guards do not affect the exhaustiveness checks, whereas the comparison operators do. Never meant to imply the type check guards would.

j14159:
No worries, just wanted to clarify :)

Think it's worth prohibiting everything but type checks in guards?  Should this just be a warning or actually an error condition?  As I said I do like the expressiveness the comparisons afford but you've shifted my thinking a bit.  If we can't rely on exhaustiveness checks or we weaken them by virtue of allowing a standard practice then I'm not sure we're actually gaining something!  Might be worth restricting things to type checks and adding `if ... then ... else`.

danabr:
Common practice (Haskell, OCaml, Scala, etc) seems to be to allow guards, and just skip exhaustiveness checking in presence of them.

Personally, I would not mind removing comparison guards. However, I'm quite sure others would disagree. I think this is one of the places where you have to state what alpaca values more: safety or expressiveness.

[...]

PR #135, opened 2017-02-13
Lambdas and one kind of binding

j14159:
Lots of changes in here, broad strokes:
- removed `#alpaca_fun_def{}`, `#var_binding{}` and `#fun_binding{}` in favour of a single binding AST node for both top-level and internal bindings, `#alpaca_binding{}`.
- functions are actual values apart from bindings.
- added the ability to define lambdas with @danabr's proposed syntax, e.g. `fn a b -> a + b`.
- for fun I added the unicode lambda and right-arrow characters (iirc @tchoutri suggested this in IRC last year)

So with this PR this is legal Alpaca:

    let use_lambda = λ x → x + 1

Or the same:

    let use_lambda = fn x -> x + 1

See the file `test_files/lambda_examples.alp` for more.

@danabr and @lepoetemaudit this touches a lot of stuff both of you have worked on and it would not at all surprise me to learn I've done a bunch of inefficient or repetitive things that should be cleaned up.  If either of you have the time to give this a look I'd be most grateful!

### DIFF
--- a/src/alpaca_parser.yrl
+++ b/src/alpaca_parser.yrl
@@ -696,25 +696,23 @@ make_define([{symbol, L, _} = Name|A], Expr, Level) ->
             case {Level, is_literal(Expr)} of
                 {top, false} -> 
                     {error, non_literal_value, Name, Expr};
-                _ -> 
-                    #alpaca_fun_def{
-                      name=Name,
-                      arity=0,
-                      versions=[#alpaca_fun_version{
-                                   line=L,
-                                   args=[],
-                                   body=Expr}]}
+                _ ->
+                    #alpaca_binding{
+                       name=Name,
+                       line=L,
+                       bound_expr=Expr}
             end;
         {ok, Args} ->
-            #alpaca_fun_def{
-               name=Name, 
-               arity=length(Args), 
-               versions=[#alpaca_fun_version{
-                            line=L,
-                            args=Args,
-                            body=Expr}]}
-%        {error, _} = E ->
-%            E
+            #alpaca_binding{
+               line=L,
+               name=Name,
+               bound_expr=#alpaca_fun{
+                             line=L,
+                             arity=length(Args), 
+                             versions=[#alpaca_fun_version{
+                                          line=L,
+                                          args=Args,
+                                          body=Expr}]}}
     end;
 make_define([BadName|Args], _Expr, _) ->
     {error, {invalid_function_name, BadName, Args}}.
@@ -778,25 +776,13 @@ all_literals([M|Rest]) ->
 
 %% Convert a nullary def into a variable binding:
 make_binding(#alpaca_fun_def{name=N, versions=[#alpaca_fun_version{args=[], body=B}]}, Expr) ->
-    #var_binding{name=N, to_bind=B, expr=Expr};
+    #alpaca_binding{name=N, bound_expr=B, body=Expr};
+%    #var_binding{name=N, to_bind=B, expr=Expr};
 make_binding(Def, Expr) ->
-    #fun_binding{def=Def, expr=Expr}.
+    Def#alpaca_binding{body=Expr}.
 
 term_line(Term) ->
-    case Term of
-        {_, L} when is_integer(L) -> L;
-        {_, L, _} when is_integer(L) -> L;
-        {bif, _, L, _, _} -> L;
-        #alpaca_apply{line=L} -> L;
-        #alpaca_cons{line=L} -> L;
-        #alpaca_map_pair{line=L} -> L;
-        #alpaca_map{line=L} -> L;
-        #alpaca_record{members=[#alpaca_record_member{line=L}|_]} -> L;
-        #alpaca_record_transform{line=L} -> L;
-        #alpaca_tuple{values=[H|_]} -> term_line(H);
-        #alpaca_type_apply{name=N} -> term_line(N);
-        #type_constructor{line=L} -> L
-    end.
+    alpaca_ast_gen:term_line(Term).
 
 add_qualifier(#alpaca_bits{}=B, {size, {int, _, I}}) ->
     B#alpaca_bits{size=I, default_sizes=false};
--- a/src/alpaca_parser.yrl
+++ b/src/alpaca_parser.yrl
@@ -454,10 +454,11 @@ infix -> term op term : make_infix('$2', '$1', '$3').
 
 literal_fun -> fn terms '->' simple_expr:
   {_, L} = '$1',
+  {ok, Args} = validate_args(L, '$2'),
   #alpaca_fun{line=L,
               arity=length('$2'),
               versions=[#alpaca_fun_version{
-                           args='$2',
+                           args=Args,
                            body='$4'}]}.
 
 %% ----- Errors (including throw, exit) --------------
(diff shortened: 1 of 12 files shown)


tchoutri:
@j14159 awww you remembered it :heart: :heart: 
Perhaps this will motivate @josevalim to do it as well ? ;)

src/alpaca_parser.yrl
```diff
@@ -449,11 +452,20 @@ tuple -> '(' tuple_list ')' :
 
 infix -> term op term : make_infix('$2', '$1', '$3').
 
+literal_fun -> fn terms '->' simple_expr:
```
danabr: Doesn't this mean you could have a lambda in the argument position of another lambda? Perhaps "term" needs to be split into terms that are safe to pattern match on and terms that are not.

To clarify:
````
let add1 = fn (fn x -> x + 1) -> 2
````
Parses, but later crashes in alpaca_codegen.

**Comment by danabr** [comment]
I'm not qualified to comment on the bulk of the changes, but I've checked out the code and played around with it. Except to the one comment I had earlier and some cleanup of unused variables this looks good to me.

File: src/alpaca_parser.yrl (hunk 1 above)
Comment (j14159) [inline src/alpaca_parser.yrl]: Good point, thanks.  We can re-use `validate_args` in the parser here with a small change to reject `#alpaca_fun{}` as a valid argument.

**Comment by j14159** [comment]
@danabr cleaned up those warnings and made functions an invalid parameter for function definition.

File: src/alpaca_parser.yrl (hunk 1 above)
Comment (danabr) [inline src/alpaca_parser.yrl]: OK. I think that we can restructure the parser in the future to make this a "syntax error" instead of "blacklisting" it.

Comment (j14159) [inline src/alpaca_parser.yrl]: Probably a good idea, yes.  Just have to break down terms, etc a little better I expect :)

**Comment by danabr** [comment]
There is one notable discrepancy/inconsistency between top level functions and lambdas that might be worth documenting, namely that top level functions may have multiple "variants"/clauses, whereas lambdas may not.

In Erlang there is no such discrepancy, since you can define lambdas like this:
````
F = fun({inc, A}) -> A + 1;
            ({dec, A}) -> A - 1
      end.
```` 

Is this something we care about to solve? I don't know what the proper syntax would look like in the case of alpaca. I'm also not sure it brings much value beyond consistency, since you can just do a match inside the lambda body if needed.

**Comment by yurrriq** [comment]
Fwiw in LFE we use `lambda` and `match-lambda`. Maybe in Alpaca we could do `fn` and `match-fn`.

**Comment by j14159** [comment]
As far as the AST is concerned we can do multiple versions of a lambda, definitely just a syntax issue.  E.g.

    type opt 'a = Some 'a | None
    let example () = (fn Some a -> a) Some 1

should work completely fine with the existing syntax, we just lack a way to express any alternative cases.  I'm not _entirely_ sure we want to contort syntax here to allow for it but I'm open to suggestions as consistency is good if we don't have to pay for it by adding cryptic bits to the language.

**Comment by OvermindDL1** [comment]
As a possibility, I was playing with this syntax in my little personal playground:
```ocaml
let singleHeadFun = fun
  a -> a in
let multiHeadFun = fun
  | (a : int) -> a + 42
  | (a : binary) -> int_of_string a
  | a -> a
in multiHeadFun (singleHeadFun 42)
```
The AST builder when checking a function to call will check each head in order to see if the arguments can fit, and if any do then it is a successful unification (the fact it is in 'order' for the type check does not matter right now but does matter for later experiments ^.^).

However the presence of `|` before each head is what marks it as a multi-head function or not, though technically you have a single head with `|` as well if you want.  The newlines are optional as always, I just like to put them for readability

j14159:
Hehe, @lepoetemaudit just raised something similar in IRC :D

I'm generally OK with this if we think it's desirable to have the short-hand for matching, my only concern is multi-argument functions.  If we require parens in each pattern then we're not really consistent with the rest of the language and if it's limited to single argument functions then it's really just short-hand for a match.

But maybe that's actually a pretty good thing for keeping things broken down and avoiding lots of nested cases?

lepoetemaudit:
Discussing it in IRC, I like the OCaml style @OvermindDL1 mentions above. I'm not against restricting it to a single argument as it's still really useful like this and it keeps the syntax consistent.

OvermindDL1:
> Discussing it in IRC, I like the OCaml style @OvermindDL1 mentions above. I'm not against restricting it to a single argument as it's still really useful like this and it keeps the syntax consistent.

Well mine does actually support multi-arg headers, but I do auto-currying on call-site, not definition-site.  :-)

It's a fun little project though, its purpose was not syntax but just typing systems, I have a bit of refined typing and all in mine.  :-)

But a multi-arg (I do not have combo let/fun in my play language like OCaml and such has, you have to assign a fun to a let to use it, the parenthesis is only for typing purposes (though I might be able to remove that restriction if I wanted...), notice it does not exist on the last one, I also show a bit of refined typing here):
```ocaml
let myAdder = fun
  | (a:int) (b:int b>=0) -> a + b
  | (a:int) (b:int) -> a - b
  | (a:binary) (b:binary) -> String.concat [a, b]
  | a b -> error "unsupported"
in myAdder 10 (-4)
```
With full typing of the fun the above would look like (I'm not typing the let part, that would be... huge, lot of heads):
```ocaml
let myAdder = fun
  | (a : int) (b : int b>=0) : int a+b -> a + b (* Since a has a valid range of infinite here, the return type is infinite+{b's range, which is {0, infinite}}, which is still infinite regardless, it is not that smart yet *)
  | (a : int) (b : int) : int a-b -> a - b (* int's extra information holds possible values allowed, simple math allowed on those values *)
  | (a : string) (b : string) : string -> String.concat [a, b] (* string's can refine on length only, currently *)
  | (a : any) (b : any) -> error "unsupported" (* any's do not have any refinement yet *)
in let curried = myAdder "Hello " (* A new erlang anonymous function is made here that curries myAdder, curried things and closures cannot be made into top-level module functions in this play language (type error) *)
in curried "world" (* Now I call it *)
```
Most things get inferred properly based on usage.  I've actually been converting my play language to try to use Elixir's AST straight (I'm losing a lot of expressiveness because of Elixir limitations sadly, but it is a fun experiment) and partly documenting it at (I'm really hating elixir's syntax, wtf `do`/`end`/`->`/`fn`/`end`/etc... inconsistencies):  https://elixirforum.com/t/mlelixir/3693

j14159:
@OvermindDL1 interesting project!  I like the refinement bits.  At some point I want to start looking at that, dependent typing, and effect typing but it's a ways off yet for me I fear :)

We could certainly allow for multi-arg heads, it would just look inconsistent with matches, e.g

    fun
      | f None   -> None
      | f Some a -> Some (f a)

is entirely doable if that's a thing we want.  My only concern is consistency in the syntax.  I think here we're also getting back to the need for guards in function heads which @danabr and I were discussing a bit in #95.  I also start to wonder with this format if we're encouraging a bit of clutter in Alpaca's syntax but I'm not sure yet.

Thoughts?

OvermindDL1:
Well `|` is what OCaml uses for function matching as well (though only on a single argument instead of multiple) so that fits (although its keyword for that is `function` instead of `fun` or `fn`, which have different meanings in OCaml ^.^).

OCaml also supports guards as well on matches with `when` as well, so that also matches.  :-)

j14159:
Opened #138 to track this and for any further discussion.

PR #136, opened 2017-02-14
Feature/error formatting

danabr:
This adds a function alpaca_error_format:fmt/2 that may be called on any error that alpaca:compile/2 produces to get a human readable error message. The intended consumer is the rebar_prv_alpaca plugin. 

I've added initial error messages for the "en-US" locale for error messages generated by the parsing and ast gen stages. These are by no means final, but rather intended as a starting point.

### DIFF
--- a/src/alpaca_ast_gen.erl
+++ b/src/alpaca_ast_gen.erl
@@ -23,9 +23,9 @@
 -type error_reason() :: {duplicate_definition, string()} |
                         {duplicate_type, string()} |
                         {function_not_exported, module(), string()} |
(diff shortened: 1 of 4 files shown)


danabr:
@j14159: Do we need tests for this? I'm unsure about what would be a good approach. One would be to compile erroneous code and check that it doesn't crash. Matching on exact error messages feels a bit lame. But perhaps it makes sense?

danabr:
After discussing with @j14159, I will close this for now until I've added gettext integration and tests.

PR #146, opened 2017-03-09
Beginning of alpaca-native integer ast nodes

j14159:
Most tests broken outside of alpaca_ast_gen.  I'm putting this commit
together so others can see roughly what I'm thinking for the purpose of
discussion and comments.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: erlang
+install: 'true'
 otp_release:
   - 19.2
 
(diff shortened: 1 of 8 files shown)


j14159:
Tests are passing now although dialyzer does fail because it can't get the abstract code for `alpaca_ast.beam`.

@danabr and @lepoetemaudit might be interested in this PR and iirc @ericbmerritt and I had discussed some aspects of the eventual AST module in the past as well.

.travis.yml
```diff
@@ -25,4 +25,6 @@ script:
   - if [ "$TRAVIS_OTP_RELEASE" != "19.2" ]; then echo  "dialyzer disabled on OTP < 19.1"; else rebar3 dialyzer; fi
   - rebar3 xref
   - rebar3 eunit
+  - rebar3 eunit
```
ypaq: Duplicate.

j14159: Intentional for now.  The first run of eunit doesn't seem to pick up Alpaca's generated tests, even if the compile phase has been explicitly done prior.

PR #148, opened 2017-03-15
RFC: Localized error messages

danabr:
Note, this is a RFC, not an ordinary pull request.

This adds a module alpaca_error_format, that can be used to get localized error messages from any  `{error, Something}` returned from `alpaca:compile/1`.

Internationalization is performed using [epo](https://github.com/brigadier/epo). In short:
* a script (`./epo scan`) is used to scan erlang source files for translatable strings, and generates a `.pot` file that may then be used as a basis for creating translation (`.po`) files. 
* `./epo compile` is then used to generate an erlang file from the translation (`.po`) files.
* Finally, at compile time, a parse transform transforms the function that is used to mark the translatable strings into a function call to the generated module.

(These steps would of course have to be spelled out and included in the documentation for a real PR).

epo was picked over the more popular (?) [gettext](https://github.com/etnt/gettext) library as it generates a single erlang module, whereas gettext is seems more aimed for server use.

Now, for the things I would like feedback on:
* Does the workflow sound reasonable? Ideally, developers would only need to add new English versions of the error messages  and regenerate the "pot" file as they add new errors; translators should only have to touch the po files; and the release manager should only need to regenerate the erlang module at release time.
* How should I test this? Currently, I only test that substitutions work. I don't know what more to test without tying the tests to a specific language and breaking them whenever the translations change.
* Is this premature? Will adding error messages just be a hassle at this point of Alpacas development?

Note that the Swedish translations are only there as an example, and would not be part of a future PR.


j14159:
I think this is a fantastic start!  @tsloughter I have a lazy question:  is there a good way to integrate the `epo` workflow @danabr has detailed into a normal build of the Alpaca compiler with rebar3?

I don't think added error messages is at all a hassle at this point in our development.  I think our current state is somewhat user-hostile and that your contribution here is a big step forward in terms of making Alpaca more accessible.  I know we're probably going to change a lot over the next year or two, hopefully as the typer is able to provide more helpful messages but I believe that making Alpaca a more...collaborative tool with things like this is key to its usefulness.

Thanks for working on this!

danabr:
I've performed the following changes:
* English is now treated as any other language.
* "epo scan" and "epo compile" are now run automatically when "rebar3 compile" is run.
* Added tests for using English as a fallback and another test that tests an actual error generated by `alpac_compile/1'.
* Added some documentation to README.md.

I'm waiting for two pull requests to be merged to epo, but otherwise I start to be quite happy with this. Let me know your thoughts and I'll squash and open a real PR.

src/alpaca_compiled_po.erl
```diff
@@ -0,0 +1,111 @@
+%%EPO COMPILED FILE
```
j14159: Should this file be committed?

j14159:
I like the ID strings you're using now.  My first reaction was that it was going to make things a little harder to manage but I think the direction you've chosen actually does make a lot of sense in that it'll reduce breakage over the long term and doesn't force us to keep maintaining "magic numbers" (the literal strings) in the code directly.

My only question is whether or not we should be including the EPO-generated module that I commented on since it's likely generated at build time like the scanner and lexer are?

Aside from that I think this is great!

yurrriq:
I'd love to translate to Spanish :)

priv/lang/alpaca.en_US.po
```diff
@@ -0,0 +1,84 @@
...
+"Invalid binary qualifier \"%(qualifier)\".\n"
+"Valid qualifiers are \\\"end\"\n"
+"-\"\\\", \\\"sign\\\", \\\"size\\\", \\\"type\\\" and \\\"unit\\\".\""
+
+msgid "invalid_bin_type %(type)"
+msgstr ""
+"Invalid binary part type \"%(type).\n"
+"Valid types are \"binary\", \"float, \"int\", and \"utf8\"."
+
+msgid "invalid_endianess %(endianess)"
```
yurrriq: `s/endianess/endianness/g`

priv/lang/alpaca.en_US.po
```diff
@@ -0,0 +1,84 @@
...
+msgid "no_module %(mod)"
+msgstr "Cannot find any module named \"%(mod)\"."
+
+msgid "type_parameter_given_to_primitive_builtin_type %(type)"
+msgstr ""
+"Type parameter provided for built in type \"%(type)\", but none were "
+"expected."
+
+msgid "unexpected_token %(token)"
+msgstr "Symtax error before \"%(token)\"."
```
yurrriq: `s/Symtax/Syntax/`

priv/lang/alpaca.en_US.po
```diff
@@ -0,0 +1,84 @@
...
+msgstr "No function \"%(fun)\" exported from module \"%(mod)\"."
+
+msgid "incomplete_expression"
+msgstr "Unexpected end of expression."
+
+msgid "invalid_bin_qualifier %(qualifier)"
+msgstr ""
+"Invalid binary qualifier \"%(qualifier)\".\n"
+"Valid qualifiers are \\\"end\"\n"
+"-\"\\\", \\\"sign\\\", \\\"size\\\", \\\"type\\\" and \\\"unit\\\".\""
```
yurrriq: Is this overly escaped?

priv/lang/alpaca.en_US.po
```diff
@@ -0,0 +1,84 @@
...
+"expected."
+
+msgid "unexpected_token %(token)"
+msgstr "Symtax error before \"%(token)\"."
+
+msgid "unknown_error %(raw_error_term)"
+msgstr ""
+"%(raw_error_term)\n"
+"Sorry, we do not have a proper message for this error yet.\n"
+"Please consider filing an issue at https://www.github.com/alpaca-lang/alpaca."
```
yurrriq: Should we append `/issues` here?

tsloughter:
Haven't had time to look at how this integrates with rebar3 yet, but I will! If I don't by Erlang Factory someone can physically poke me until I do :)

yurrriq:
I've made a PR on top of this one with some nitpicking and an `alpaca.es_MX.po`: https://github.com/danabr/mlfe/pull/1

priv/lang/alpaca.en_US.po
danabr: Right. Done.

priv/lang/alpaca.en_US.po
danabr: Fixed. Thanks!

priv/lang/alpaca.en_US.po
danabr: Yes, fixed.

priv/lang/alpaca.en_US.po
danabr: Funny I managed to get this wrong both in po file *and* in the unit tests. Copy-paste error, I guess.

src/alpaca_compiled_po.erl
danabr: Not any longer, since we automatically generate it. I've removed it and added it to .gitignore.

danabr:
@yurrriq: Once I get an approval for this, I'll squash this and then you can open your PR directly towards alpaca.

yurrriq:
@danabr at least you were consistent :laughing:. I'll amend my PR to just include the Spanish stuff and wait to open it on Alpaca itself.

yurrriq:
@danabr here's a patch for what I'd still change in `alpaca.en_US.po`:

```diff
From 1da1e62308de1f4fdaba8c527111e327b064ad62 Mon Sep 17 00:00:00 2001
From: Eric Bailey <eric@ericb.me>
Date: Sat, 18 Mar 2017 19:26:21 -0500
Subject: [PATCH] Update alpaca.en_US.po

Fix a couple cases of bad quoting/spacing.
---
 priv/lang/alpaca.en_US.po | 16 ++--------------
 1 file changed, 2 insertions(+), 14 deletions(-)

diff --git a/priv/lang/alpaca.en_US.po b/priv/lang/alpaca.en_US.po
index 04ae573..f304001 100644
--- a/priv/lang/alpaca.en_US.po
+++ b/priv/lang/alpaca.en_US.po
@@ -36,12 +36,12 @@ msgstr ""
 
 msgid "invalid_bin_type %(type)"
 msgstr ""
-"Invalid binary part type \"%(type).\n"
+"Invalid binary part type \"%(type)\".\n"
 "Valid types are \"binary\", \"float, \"int\", and \"utf8\"."
 
 msgid "invalid_endianness %(endianness)"
 msgstr ""
-"Invalid endianness \"%(endianness)\".Did you mean \"big\", \"little\", or "
+"Invalid endianness \"%(endianness)\". Did you mean \"big\", \"little\", or "
 "\"native\"?"
 
 msgid "invalid_fun_parameter"
@@ -75,15 +75,3 @@ msgstr ""
 "Sorry, we do not have a proper message for this error yet.\n"
 "Please consider filing an issue at https://www.github.com/alpaca-lang/alpaca/"
 "issues."
-
-#~ msgid "invalid_endianess %(endianess)"
-#~ msgstr ""
-#~ "Invalid endianness \"%(endianness)\".Did you mean \"big\", \"little\", or "
-#~ "\"native\"?"
-
-#~ msgid ""
-#~ "%(raw_error_term)\n"
-#~ "Sorry, we do not have a proper message for this error yet.\n"
-#~ "Please consider filing an issue at https://www.github.com/alpaca-lang/"
-#~ "alpaca."
-#~ msgstr "Redefintion of module name from \"%(old)\" to \"%(new)\"."
-- 
2.12.0
```

danabr:
Squashed all my changes into a single commit, and put @yurrriq's patch on top.

@j14159: Unless you or others have anything more you want to see added, this is ready for merge.

PR #149, opened 2017-03-19
Native symbols

j14159:
This changes every internal use of the `symbol` type to that of the Alpaca-native `Symbol` AST node (see `src/ast.alp`).

### DIFF
--- a/src/alpaca_ast_gen.erl
+++ b/src/alpaca_ast_gen.erl
@@ -132,7 +132,8 @@ rename_and_resolve(Modules) ->
 expand_exports(Modules) ->
     expand_exports(Modules, []).
 
-name_and_arity(#alpaca_binding{name={symbol, _, N}, bound_expr=E}) ->
+name_and_arity(#alpaca_binding{name=BindingName, bound_expr=E}) ->
+    N = alpaca_ast:symbol_name(BindingName),
     case E of
         #alpaca_fun{arity=A} -> {N, A};
         _                    -> {N, 0}
@@ -143,7 +144,7 @@ expand_exports([], Memo) ->
 expand_exports([M|Tail], Memo) ->
     F = fun({_, _}=FunWithArity) -> 
                 FunWithArity;
-           (Name) when is_list(Name) ->
+           (Name) when is_binary(Name) ->
                 NameAndArity = [name_and_arity(F) || F <- M#alpaca_module.functions],
                 [{Name, A} || {N, A} <- NameAndArity, N =:= Name]
         end,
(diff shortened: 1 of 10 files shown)


src/ast.alp
```diff
@@ -1,18 +1,55 @@
...
 
-type ast = Int {line: int, val: int}
-         | Float {line: int, val: float}
-         | String {line: int, val: string}
+type opt 'a = Some 'a | None
+
+type typ = TInt
+         | TFloat
+         | TString
+         | Arrow (list typ, typ)
```
danabr: TArrow for consistency?

src/ast.alp
```diff
@@ -22,4 +59,32 @@ let float line val = Float {line=line, val=val}
...
+
+let bind name line bound body =
+  Binding {  line=line
+           , name=name
+           , bound=bound
+           , body=Some body
+           , typ=None
+          }
+
+let bind_body Binding b_rec body = Binding {body=Some body | b_rec}
```
danabr: Although I know the parse rules, I still find this confusing when quickly scanning through the code. I either intepret it as:
* a function taking three arguments (Binding, b_rec and body)
* or a function taking one argument, and pattern matching against a two argument constructor.

Only the knowledge that there is no such thing as a two argument constructor, and that Binding takes a type argument, disambiguate this so I can parse it as what it actually means, namely: 
`let bind_body (Binding b_rec) body ...`

I wonder if we shouldn't try to encourage this less ambiguous style.

danabr:
Looking forward to the day when we can re-implement the exhaustiveness checker in pure alpaca!

src/alpaca_ast_gen.erl
```diff
@@ -437,13 +441,15 @@ rename_bindings(Environment, #alpaca_binding{}=TopLevel) ->
...
 
 rebind_args(#env{current_module=Mod}=Env, Map, Args) ->
-    F = fun({symbol, L, N}, {#env{next_var=NV}=E, AccMap, Syms}) ->
+    F = fun({'Symbol', _}=S, {#env{next_var=NV}=E, AccMap, Syms}) ->
+                N = alpaca_ast:symbol_name(S),
+                L = alpaca_ast:line(S),
                 case maps:get(N, AccMap, undefined) of
                     undefined ->
-                        Synth = next_var(NV),
+                        Synth = unicode:characters_to_binary(next_var(NV), utf8),
```
danabr: You could move the list -> binary translation into next_var, since all callers of next_var now make that conversion.

src/ast.alp
j14159: This is a good point and I think there's likely value in letting the (eventual) formatter clean this sort of ambiguity up to enforce the style you're suggesting.

src/ast.alp
j14159: Yes :)  Thanks for pointing that out.

src/alpaca_ast_gen.erl
j14159: Good point, fixing.

j14159:
Declining and resubmitting for rebase.

PR #151, opened 2017-03-20
Add internationalization of error messages.

danabr:
This adds a module alpaca_error_format, that can be used to get localized
error messages from any `{error, Something}` returned from `alpaca:compile/1`.
    
Translations are stored under priv/lang. See priv/lang/alpaca.en_US.po
for an example.
    
Whenever alpaca is compiled, the [epo](https://github.com/brigadier/epo)
tool is run to generate a .pot file from all strings wrapped in `__()`.
The same tool also turns .po files into an erlang module containg all
translated messages.


danabr:
@j14159: Closed the RFC, and opened this one as a proper PR.

yurrriq:
Vad hände med den svenska översättningen?

danabr:
I never intended to include the Swedish translation. I only used it as an example. Now that we use message ids, the English translations serve that purpose.

j14159:
@danabr thanks for this, love it!

PR #155, opened 2017-03-22
Allows forward references to top-level bindings

### LINKED ISSUE #154: Forward references that aren't applications fail typing
Thanks to @lepoetemaudit for reporting this one in IRC:

    module forward_symbol_reference

    let apply f x = f x

    let hof_fail () =
        apply add10 5

    let val_fail () =
        10 + val

    -- As these are both declared AFTER their usage, neither
    -- of the above functions will compile

    let add10 x = x + 10
    let val = 5

Basic function application will look forwards in the module for functions declared later but symbol lookup doesn't have this same fallback (but should).

j14159:
Fixes #154.  If there are multiple versions of the top-level binding
referred to this picks the first occurrence since a simple symbol
reference can't tell us which arity we should be looking for (vs
function application where we can make reasonable decisions).

### DIFF
--- a/src/alpaca_typer.erl
+++ b/src/alpaca_typer.erl
@@ -1261,9 +1261,38 @@ inst_binding(VarName, Line, Lvl, #env{bindings=Bs} = Env) ->
     Default = {error, {bad_variable_name, module_name(Env), Line, VarName}},
     case proplists:get_value(VarName, Bs, Default) of
         {error, _} = E ->
-            E;
+            case lookup_binding(VarName, Env, E) of
+                {ok, T} -> inst(T, Lvl, Env, maps:new());
+                Err     -> Err
+            end;
         T ->
             inst(T, Lvl, Env, maps:new())
+    end.
+
+%% If inst_binding/4 can't find a binding in the environment for the name it was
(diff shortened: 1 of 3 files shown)


lepoetemaudit:
Ah hah so that was it! Thanks so much for fixing this :)

src/alpaca_typer.erl
```diff
@@ -1261,11 +1261,40 @@ inst_binding(VarName, Line, Lvl, #env{bindings=Bs} = Env) ->
...
+            MatchingFuns = [B || #alpaca_binding{
+                                    name={'Symbol', #{name := N}}}=B <- Funs,
+                                 N =:= VarName
+                           ],
+            case MatchingFuns of
+                [] ->
+                    Default;
+                %% There's no way to tell which arity version of a function is
+                %% intended from a symbol alone so we're defaulting to the
+                %% first one bound to the name we're looking for:
```
danabr: I guess this is yet another reason to avoid functions with the same name but different arity.

j14159: Those reasons do seem to be piling up a bit :)

PR #157, opened 2017-03-22
Rewrite/bind lambdas in tests

### LINKED ISSUE #143: Lambdas don't work in tests
Rewriting of lambdas isn't applied to the bodies of tests so the following will result in a compiler error:

    test "mapping a doubling lambda on an option doubles the contents" =
      assert.equal Some 2 (map (fn x -> x + x) Some 1)

j14159:
Fixes #143, allows us to use lambdas in the body of tests, specifically
when passed as arguments to other functions under test.

### DIFF
--- a/src/alpaca_codegen.erl
+++ b/src/alpaca_codegen.erl
@@ -251,9 +251,7 @@ gen_fun_version(Env, #alpaca_fun_version{args=Args, guards=Gs, body=Body}) ->
     end.
 
 gen_tests(Env, Tests) ->
-    io:format("~p~n", [Tests]),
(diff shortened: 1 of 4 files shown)


src/alpaca_codegen.erl
```diff
@@ -240,7 +251,10 @@ gen_fun_version(Env, #alpaca_fun_version{args=Args, guards=Gs, body=Body}) ->
     end.
 
 gen_tests(Env, Tests) ->
-    gen_tests(Env, Tests, []).
+    io:format("~p~n", [Tests]),
```
danabr: Intentional to leave these printouts?

j14159: Oh good catch, thanks!

PR #160, opened 2017-03-25
Fix mistake in usage example

monkeygroover:
- the shell usage example did not include the alpaca_ in the module name.


j14159:
Thanks for catching this!

PR #161, opened 2017-03-25
Escape \ style sequences in strings

lepoetemaudit:
I thought it most appropriate to perform this in the scanner. This only handles the simpler cases as described here: http://erlang.org/doc/reference_manual/data_types.html#id76758, ignoring hex, octal and Control A to control Z, but it would be possible to add these if they were desirable and didn't interfere with UTF8 rules. This also doesn't escape single quotes as I'm not sure there's any necessity.


j14159:
Thanks!  Guessing the absence of this made a few things...difficult ;)

PR #167, opened 2017-04-20
Beam loading

lepoetemaudit:
This PR allows providing precompiled BEAM files to the Alpaca compiler, so that it is not necessary to recompile every file (including from every dependency) on every compilation. The type information is stored (stripped of function bodies) as a module attribute, meaning it can be retrieved by the Alpaca compiler. To aid external users of the compiler (such as the rebar3 plugin, Elixir mix compiler) a new function, `hash_source/1` is exposed, which returns a simple hash of the source. This hash is also stored with generated BEAM files as a module annotation, so external users can see if a given source file is 'dirty' in comparison to a BEAM file (i.e. has the Alpaca version changed since compilation, and/or has the source changed).

In addition, this PR also contains the beginning of a `compiler_info` function on the `alpaca` module, return a property list. This currently only contains `version`, used by the source hashing described above.


lepoetemaudit:
```
===> /home/travis/build/alpaca-lang/alpaca/_build/default/plugins/alpaca/src/alpaca_ast_gen.erl:61: Warning: alpaca_ast_gen:make_modules/1 is unused export (Xref)
```

I think this is causing Travis to show the build as failed, I'll look into it.

j14159:
Looking good!  Thanks for this!

Might be helpful to include a function that retrieves the hash for a `.beam` file given its name in `alpaca.erl`, that way the rebar3 plugin (or whatever tool) can just use that function and `hash_source/1` in order to make a quick and simple decision about what to pass in.  What do you think?

lepoetemaudit:
Good Idea, I'll add a `retrieve_hash/1` function that takes a file path.

lepoetemaudit:
Spotted a couple of little cosmetic things that I've now fixed. @j14159 let me know if there are any other functions you can think of that might be useful to external tools at this time.

src/alpaca.erl
```diff
@@ -28,6 +30,9 @@
              , compile/2
              , file/1
              , file/2
+             , compiler_info/0
+             , hash_source/1
+             , retrieve_hash/1
```
j14159: I think we want to export this too?

src/alpaca.erl
```diff
@@ -36,13 +41,19 @@
 -include_lib("eunit/include/eunit.hrl").
 -endif.
 
+-define(COMPILER_VERSION, "0.2.7").
```
j14159: Just thought, we might want to already call this `"0.2.8"` as that's the next release and `0.2.7` is already tagged.

src/alpaca.erl
lepoetemaudit: Ah whoops, that might help!

src/alpaca.erl
lepoetemaudit: Good point, I'll amend.

j14159:
Thanks for this!

PR #169, opened 2017-05-07
Module deps

lepoetemaudit:
Adds a function for listing out dependencies from Alpaca source without having to compile.

This is useful for calculating the dependency graph of a given module, meaning that an external tool such as the rebar plugin may feed modules into Alpaca in the optimal order, and may force a module to be recompiled if one of its dependency changes.

I put this into the `alpaca_ast_gen` module as it was able to reuse some of the mechanisms there, but perhaps the list_dependencies fun should also be exposed from the alpaca main module for the 'public interface' aspect.

### DIFF
--- a/src/alpaca.erl
+++ b/src/alpaca.erl
@@ -23,6 +23,7 @@
         , compiler_info/0
         , hash_source/1
         , retrieve_hash/1
+        , list_dependencies/1
         ]).
 
 %% Can be safely ignored, it is meant to be called by external OTP-apps and part
@@ -34,6 +35,7 @@
              , compiler_info/0
              , hash_source/1
              , retrieve_hash/1
+             , list_dependencies/1
              ]).
 
 -include("alpaca_ast.hrl").
@@ -109,6 +111,9 @@ hash_source(Src) ->
 retrieve_hash(Filename) ->
     {ok,{_,[{attributes,A}]}} = beam_lib:chunks(Filename,[attributes]),
     proplists:get_value(alpaca_hash, A).
+
+list_dependencies(Src) ->
+    alpaca_ast_gen:list_dependencies(Src).
 
 maybe_print_exhaustivess_warnings(Warnings, Opts) ->
   case proplists:get_value(warn_exhaustiveness, Opts, true) of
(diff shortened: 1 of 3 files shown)


lepoetemaudit:
Thinking about it, it might be also be useful for this fun to return a tuple of {ModuleName, Dependencies} so that an external tool doesn't need to rely on the file name to figure out a given module's name. Any thoughts?

j14159:
I suppose that would probably make it a little simpler on the caller's side, yeah.

Checking my understanding on where this fits in with the compilation process:
- rebar3 plugin (or any other tool using Alpaca) gets a list of source and `.beam` files
- plugin checks hashes on each `.beam` vs source, discards out of date `.beams`
- plugin checks dependencies for each source file to compile, ordering appropriately

Mostly just trying to make sure I'm not oversimplifying or missing anything :)

lepoetemaudit:
@j14159 that's exactly right. The only other detail is that by knowing what modules a module depends on, it can also be marked as being 'out of date' if one of those dependencies (and so on, recursively) needs to be recompiled.

I'll make it return the module name as well (if included). It's meant to be a 'best possible' pass, I'd rather leave error checking for the actual compiler!

lepoetemaudit:
The function now returns the tuple of {ModuleName, Dependencies}, and it's also now exposed from the alpaca module so that it's consistent with the other utility / compiler entrypoint functions.

j14159:
Oh, minor thing:  mind adding a quick comment to the method exported from `alpaca.erl` just so it's obvious later for others as to what it's supposed to do?

This is all awesome by the way, thanks for running with this!

lepoetemaudit:
Yes, will definitely add that comment.

I'm definitely excited to get the rebar plugin pulling a release and building things nicely :)

lepoetemaudit:
I've commented on that function, and the hashing functions too: I also found and fixed a little bug around function imports that would break list_dependencies. Also tweaked the formatter a bit. I have a branch of the rebar plugin that uses the error formatting.

I've introduced some minor whitespace changes from my editor but they can be ignored with this link: https://github.com/alpaca-lang/alpaca/pull/169/files?w=1

j14159:
Looks good, thanks for this!  Out of curiosity have you done any work on locale detection for the rebar3 plugin yet?  I did some initial cursory looking a little while ago but didn't come up with anything obvious.

lepoetemaudit:
Thanks! Amongst other things, this PR https://github.com/alpaca-lang/rebar_prv_alpaca/pull/12 has locale detection - it just tries to read and parse `$LANG` from the environment, defaulting to en_US if not found. I've been enjoying error messages in es_MX of late :)

Issue #170, opened 2017-05-14
Synthetic type variable names crash exhaustiveness checks


j14159:
Ran into this while trying to add more native AST nodes, e.g. given

    type opt 'a = Some 'a | None

    type module_ast = Module {line: int, name: atom}
                    | Exports list {line: int, name: string, arity: opt int}

    let make_export l n = {line=l, name=n, arity=None}
    let make_export l n a = {line=l, name=n, arity=Some a}

Adding the following method breaks the exhaustiveness checker because `lists:key_find/3` can't find a type variable:

    let make_exports es = Exports es

Breaks because the parser synthesizes a type variable name to bind `int` to when it encounters `opt int` in `Export`'s record above.  The exhaustiveness checker uses the actual definition of the type for exhaustiveness checks because using the complete list of type members is the only sensible way to do it but the synthesized variable names do not match the type's real variable names.

These also cause the same variable lookup failure:

    type r = R list {x: int, y: opt int}

    let make_r r = R r

    type t = T map string (opt int)

    let make_t x = T #{"x" => x}

    type v = V list (opt int)
 
    let make_v vs = V vs

While these do not:

    type u = U opt int

    let make_u x = U Some x

    type int_opt = opt int

    type x = X int_opt

    let make_x x = X Some x

    type z = Z (int, opt int)

    let make_z z = Z (z, Some z)

I suspect we need to do one of the following:
- as part of AST generation, rewrite types to replace synthetic type variable names with the correct ones.  Insufficient information at this point is an error on the part of the user anyhow although this does likely duplicate some of the work the typer does.
- let the typer rewrite the types.  I'm not certain this is really the best place to do AST manipulation though as the only modifications the typer should really be making are annotating expressions with their inferred types.
- rewrite ADT variables to the correct ones in the exhaustiveness checker.  Tempting to just patch it here but I don't think this should really be the responsibility of the exhaustiveness checker.  It should be able to trust that the ADT it gets is correct.

I'm leaning towards the first option above but there's a catch in that a module may refer to _another_ module's ADTs so this operation needs to be able to inspect all types available to the program.  This suggests there's a new distinct stage.

PR #171, opened 2017-05-18
Default imports

lepoetemaudit:
Submitting this PR so we can discuss.

This adds an option to existing opts list in alpaca:compile of `default_imports` which allows the user to specify functions and types that will be brought into scope in every module without explicit imports.

The main aim of this is so that we can have 'standard library' functions and types that are available to all modules. I imagine that alpaca_lib could define its defaults in a file which the rebar plugin could read; if alpaca_lib is available to the rebar plugin, it would use this list of defaults. This would mean that the default functions could actually be changed by alpaca_lib itself as more are added. Some candidates would be operators such as (<|), (|>), and types such as `maybe/some` and `result`, and general utility functions like `always`, `identity`.

With this approach (explicitly listing all imports and types) we don't need to bundle all defaults into a single module, but they can be specified quite naturally across multiple modules.

I originally used Alpaca's own AST representation of the imports, but I decided to make them a bit more consistent, so function imports are defined as `{Mod, Fun}` or `{Mod, Fun, Arity}` and types are specified as `{Mod, Type}`.

### DIFF
--- a/src/alpaca.erl
+++ b/src/alpaca.erl
@@ -83,7 +83,9 @@ compile_phase_1(Sources, Opts) ->
                             filename:extension(FN) == ".beam"
                         end, Sources),
 
-    case alpaca_ast_gen:make_modules(AlpacaSrcs, BeamFiles) of
+    DefaultImports = proplists:get_value(default_imports, Opts, {[], []}),
+
+    case alpaca_ast_gen:make_modules(AlpacaSrcs, BeamFiles, DefaultImports) of
         {error, _}=Err -> Err;
         {ok, Mods} ->
             compile_phase_2(Mods, Opts)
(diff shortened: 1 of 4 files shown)


j14159:
Back from vacation now so should be a bit more responsive :)

I think this is a great step, thanks for working on this!  Two things:
- seems to me like we might want to get Alpaca's AST more in line with what you're doing here for imports, using a standard `{Module, Function, Arity}` tuple.  The varying formats that exist now are really only there to make parsing of subsets/lists of functions in a module more simple to communicate to the AST gen and typer but there's really no reason the AST generation stage shouldn't be making this more intelligible to all down-stream consumers by expanding everything to a proper `list({Module, Function, Arity})`.  Similar theme to #170 I think.  Definitely *not* suggesting that should be part of this work, just that maybe we want the consistency sometime sooner than later (v0.2.8 or v0.2.9 or maybe whenever we move imports/exports to the Alpaca-native AST?)
- I think the idea of a library specifying default imports is really interesting but we might want to have a slightly larger conversation about whether we're privileging alpaca_lib or introducing a feature that, however potentially useful, could cause some headaches when libraries conflict later.  Maybe there's a happy medium?

Thoughts?

lepoetemaudit:
Great to have you back :)

I definitely agree with making the AST representation of imports more consistent and obvious. I also note that type imports are a record type, whereas function imports are bare tuples. Perhaps indeed a good candidate for moving to the Alpaca native AST types.

I agree with the worry about privileging alpaca_lib. I kept the compiler itself entirely neutral in this PR. My thinking was that the user could specify the default imports in some configuration somewhere, but that perhaps if alpaca_lib was specified as a dependency, the rebar plugin would use its imports by default (those imports specified in some file in alpaca_lib). If a user wanted to override those imports (or add to them) they'd just need to provide the right configuration.

My desire is to have sensible defaults which are easy to override. I certainly would imagine people wanting to use alpaca_lib _most_ of the time. It's normal in OCaml to replace the standard library with Core or Batteries, but that is only because the built-in stdlib is so poor and unfortunately having competing stdlibs has split the community. Having said that, I'd want alpaca_lib to be pretty small, but to be the main dependency for most other useful libraries.

j14159:
Yeah probably a good idea to move both type and function imports to the native AST as soon as possible but might need to hold off until 0.2.8 because of #170.  I'll dig in once I squash that bug.

I guess leaving the default imports entirely to the build tool and its configuration is a relatively straightforward thing to do - I think we're basically in agreement with everything here including the compact stdlib.  The "kitchen sink" approach in Scala has bothered me on occasion :)

j14159:
Also thanks again :)

PR #172, opened 2017-05-21
Allow (<) leading infixes to have right assoc

lepoetemaudit:
This PR causes infixes that start with (<) to be right associative.

Allowing infixes to be left or right allows defining an operator such as Elm's pipe backward/apply `(<|)` , similar to Haskell's `($)`, which is really useful for eliminating parentheses, e.g. 

`add <| add <| add 10` instead of `add (add (add 10))`

With Yecc the associativity has to be predefined, i.e. it wouldn't be possible to specify the desired associativity in Alpaca source code. OCaml's approach is that the leading char in the infix specifies its associativity. In this PR everything is currently assumed to be left assoc, unless it begins with a leading `<`. The associativity is defined here: https://github.com/ocaml/ocaml/blob/trunk/parsing/parser.mly#L563 and the operator groups here: https://github.com/ocaml/ocaml/blob/trunk/parsing/lexer.mll#L500

For context, I'm thinking of operators that might be defined by standard in alpaca_lib.

### DIFF
--- a/src/alpaca.erl
+++ b/src/alpaca.erl
@@ -23,7 +23,7 @@
         , compiler_info/0
         , hash_source/1
         , retrieve_hash/1
-        , list_dependencies/1
+        , list_dependencies/1 
         ]).
 
 %% Can be safely ignored, it is meant to be called by external OTP-apps and part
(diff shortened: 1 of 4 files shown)


j14159:
Makes sense to me, thanks for this!

PR #177, opened 2017-06-04
Allow currying on local 'let' bound funs

lepoetemaudit:
I realised I never got round to this - it turned out to be much easier than I thought (logically, if the typer finds an arrow that it can unify, and the arrow didn't come from the top level or another module, it must be from a let binding).

I guess the only concern is whether this might interfere with defining multiple arity funs in local bindings, but AFAIK it is currently impossible to define multiple let bindings with differing arities and the same name anyway; if it became possible in future we could apply the same rules for detecting ambiguous curries.

If we're happy about the currying rules overall, this also closes issue https://github.com/alpaca-lang/alpaca/issues/74

### DIFF
--- a/src/alpaca_typer.erl
+++ b/src/alpaca_typer.erl
@@ -1908,6 +1908,27 @@ typ_of(Env, Lvl, #alpaca_apply{line=L, expr=Expr, args=Args}) ->
     %% to find a potential function that can be unambigiously
     %% curried, and then types against that by manipulating the
     %% argument list and return type
+    LocalCurryFun =
+        fun() ->
+            %% If we got an arrow type, if we couldn't find it in the
+            %% top level, and it still didn't unify, it might be
+            %% a local binding we can still curry.
+            case typ_of(Env, Lvl, Expr) of
+                {error, {bad_variable_name, _, _, _}} = E -> E;
+                {error, _} = E -> E;
+                {{t_arrow, TArgs, TRet}, NextVar} ->
+                    case length(Args) >= length(TArgs) of
+                        true -> 
+                            {'Symbol', #{name := N, line := Ln}} = Expr,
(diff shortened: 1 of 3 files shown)


src/alpaca_typer.erl
```diff
@@ -4760,19 +4781,7 @@ different_arity_test_() ->
               ?assertMatch(
                  {error, {not_found, _, <<"add">>, 2}},
                  module_typ_and_parse(Code))
-      end
-    , fun() ->
```
j14159: Any particular reason we're removing this test case?

lepoetemaudit: Yes - the test no longer succeeded as `f x` now becomes a valid currying of `f`, meaning that it would return an error as previously expected. I tested that the currying works correctly elsewhere though. I believe there's also another test which checks than an error is returned when too many arguments are passed. Would it be preferable to keep this test as well and pass too many arguments here, perhaps? Or change the assert to check that it returns without error?

j14159: Ah, of course, I should have thought through this a bit more.  Thanks for the explanation.

j14159:
Thanks for this, I like the consistency :)

I'm not particularly bothered about having multiple arity funs work inside local bindings to be perfectly honest.  I'm sure someone has a reasonable argument for them somewhere but it seems a little overkill to me.  We only support multiple arity at top-level to be friendly with Erlang behaviours anyhow.

yurrriq:
This is great. Thanks, @lepoetemaudit!

lepoetemaudit:
@yurrriq no problem :) I hit this again when working on the REPL as I needed to bring all expressions and binds into function scope there.

@j14159 I agree, I don't feel multiple arity funs are particularly useful anyway outside the top level.

PR #181, opened 2017-06-14
[WIP] Nicer errros

Licenser:
WIP for nicer errors.

**Work In Progress don't merge!**

Goals:

- [x] Include source where possible.
- [x] Highlight syntax errors.
- [x] Do some general pretty coloury thingies.
- [x] Do some pretty coloury thingies for messages themsefls.
- [x] Ensure indenting
The schema / rules used is the following:

We avoid bold colours as they can seem intimidating, with the exception of parts we **really** want people to see.

* Filenames are underlined.
* Module are cyan
* line number are cyan with the exception of the errors line number in code snippets
* Errors are red
* 'Good' data is green.
* Important information that is neither error not 'good' is bold.
* Filename/line-number are in one line separated by a `:` so editors can understand that.
* Error messages are on the following line, indented by two spaces
* Code is indented 6 spaces (or rather the number is 4 spaces 'wide' and is always indented 2 more spaces)
* Whenever possible the code snipped is included in the error with the two surrounding lines.

Or in an example (please forgive the horrible yellow arrows):
image

1) the module name part of the file.
2) the line number the error occurred in
3) both lines are indented by 2 spaces
4) We're dealing with a PID, this is important info but not an error
5) We would have liked arity 1
6) We got arity 42

Code is included and on syntax errors the offending syntax is highlighted:
image

### DIFF
--- a/priv/lang/alpaca.de_DE.po
+++ b/priv/lang/alpaca.de_DE.po
@@ -12,7 +12,7 @@ msgstr ""
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 "X-Generator: Poedit 1.8.7.1\n"
 
-msgid "builtin_type_arity_error %(num_expected) %(num_supplied)"
+msgid "builtin_type_arity_error %(type) %(num_expected) %(num_supplied)"
 msgstr ""
 "Falsche Anzahl an Parametern für eingebauten Typ\"%(type)\".\n"
 "Erwartet %(num_expected), erhalten %(num_supplied)."
@@ -23,7 +23,7 @@ msgstr "Mehrfache Definition von \"%(id)\"."
 msgid "duplicate_type_definition %(id)"
 msgstr "Der Typ \"%(id)\" wurde bereits definiert."
 
-msgid "function_not_exported %(mod) %(name)"
+msgid "function_not_exported %(mod) %(fun)"
 msgstr "Das Modul \"%(mod)\" exportiert die Funktion \"%(fun)\" nicht."
 
 msgid "incomplete_expression"
--- a/priv/lang/alpaca.en_US.po
+++ b/priv/lang/alpaca.en_US.po
@@ -12,7 +12,7 @@ msgstr ""
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 "X-Generator: Poedit 1.8.7.1\n"
 
-msgid "builtin_type_arity_error %(num_expected) %(num_supplied)"
+msgid "builtin_type_arity_error %(type) %(num_expected) %(num_supplied)"
 msgstr ""
 "Wrong number of type parameters provided for builtin type \"%(type)\".\n"
 "Expected %(num_expected), but got %(num_supplied)."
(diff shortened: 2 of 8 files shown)


Licenser:
First glance: 

image

j14159:
I love this but have a question:  error/good is red/green and this might be an accessibility issue for some users.  Can we pick a pairing that will be usable for those with colour blindness?  Cursory searching turned up a few helpful links:

http://www.visualisingdata.com/2015/11/colour-swatch-alternatives-to-green-and-red/
http://www.somersault1824.com/tips-for-designing-scientific-figures-for-color-blind-readers/

Open to other suggestions, including adding some symbols with the colours (h/t @saem).

yurrriq:
`#4dac26` / `#d01c8b` looks nice to me.

Licenser:
I'm not sure how the status is on outputting abritary colors on the console but I'll take a look on if/how that can be done. FWIW it will fallback gracefully for dumb terminals that don't do collors and return to the good old grey no white

Licenser:
Okay it's possible, you never stop learning ;) I'll give it a crack to make `cf` compatible with truecolor and then update the PR.

Licenser:
Works:
image

j14159:
Thanks for this, and for the accessibility win!

NobbZ:
when you stick to the 8 base colors and their bold versions, colorblind
people can configure their terminal emulator to show the colors they want,
but by forcing 256 color mode this isn't possible any more.

Jeremy Pierre <notifications@github.com> schrieb am Mi., 14. Juni 2017,
22:39:

j14159:
@NobbZ thanks, did not realize!  @Licenser thoughts?

Licenser:
That's a very good point, I never thought about that either, I'll ask around on twitter getting some idea what people affected like better might help be the right path?

j14159:
I'm happy to see/hear more about what people want.

sanmiguel:
Since @Licenser asked for input on twitter, I came here to give my opinion but it looks like @NobbZ already nailed it: if you stick to the base term colours, I can configure my terminal to suit me (I have red/green deuteranomaly). But I think that this goes beyond just inclusion for folks with some form of colour blindness.

Some more background (ha ha):
Personally I use the excellent [solarized](http://ethanschoonover.com/solarized) colourscheme everywhere, normally with a dark background. But if it's bright out, I switch to a light background.
All your screenshots are on a black background: how will this output render on a white background? Or solarized-base3 `#fdf6e3` ? Or darkblue? If you stick to standard term colours, it's mostly taken care of by the term. But if you go the route of 256-colour, you'll need to either make sure your colours work on all background colours, or provide configurable profiles. Or I guess simply accept that people turn off colours if they're on a non-dark background...

Just my $0.02...

On another note, I'm super pleased to see consideration for how this output works with tooling. It is seriously frustrating when tooling makes it hard to parse output via an `errorformat` in vim. Mentioning no names **cough** _eunit_ **cough**  _common-test_ **cough**

Licenser:
Sounds like base colours is the better approach then :) I'll make a new PR for that.

on the topic of light backgrounds, I kind of suspect/expect people who change the background to adjust the base colours to something that works on the background as it possibly affects a lot of tooling. Same way that it switches the neutral color from white to black (or light grey to dark grey?) to keep the text readable.

Issue #97, opened 2017-01-17
Boolean operator support


danabr:
As far as I can tell, we cannot do things like ``(x == 1) || (y < 4)`` and ``(x == 1) && (y < 4)``.

Licenser:
I've put in some logic in #182 but I think it's a bit wring given it does not short curcuite. It might be better to implement `||` and `&&` in something like core-alpaca if that exists?

yurrriq:
Perhaps `andalso` and `orelse` are more appropriate.

Licenser:
That is what I thought too, but it turns those are not functions in erlang. When I thought about it, it made sense as you can't implement short circuting and/or with a function.

Basically I think there are three options:
1) don't short circuit - as done in #182 
2) generate the code for them as part of the code generation (i.e. turn `||` into something like `case Left of true -> true; false -> Right end`)
3) since alpaca can define infix operators, define it in alpaca itself

Personally 3 feels like the best option but I'm not sure if that has implications that I don't see.

yurrriq:
Ah, right. Option 3 sounds good to me.

Licenser:
Well I'm stupid, implementing them as infix functions will have the same problem as `:and` and `:or` -.-

j14159:
@Licenser nobody is stupid, just at various points in figuring out or learning what to do :)

j14159:
(e.g. I have no idea here at present)

yurrriq:
Perhaps the `case` idea is solid. Later I'd like like to examine the Core Erlang generated by something like this:

```erlang
'&&'(A, B) ->
  A andalso B.
```

j14159:
Closing per PR #182 for now.

yurrriq:
#### `foo.erl`

```erlang
-module(foo).

-export(['&&'/2]).

'&&'(A, B) ->
    A andalso B.
```

#### `foo.core`

```fish
erlc +to_core foo.erl
```

```erlang
module 'foo' ['&&'/2,
	      'module_info'/0,
	      'module_info'/1]
    attributes []
'&&'/2 =
    %% Line 5
    fun (_cor1,_cor0) ->
	%% Line 6
	( case _cor1 of
	    ( <( 'true'
		 -| ['compiler_generated'] )> when 'true' ->
		  _cor0
	      -| ['compiler_generated'] )
	    ( <( 'false'
		 -| ['compiler_generated'] )> when 'true' ->
		  'false'
[...]

PR #182, opened 2017-06-14
Add && and ||

Licenser:
This adds `||` and `&&` linking to `erlang:or` and `erlang:and` related to #97

### DIFF
--- a/src/alpaca_scan.xrl
+++ b/src/alpaca_scan.xrl
@@ -132,10 +132,11 @@ c"(\\"*|\\.|[^"\\])*" :
 \+ : {token, {plus, TokenLine}}.
 
 
-
 [\*\/\%]   : {token, {int_math, TokenLine, TokenChars}}.
 {FLOAT_MATH} : {token, {float_math, TokenLine, TokenChars}}.
 ->       : {token, {'->', TokenLine}}.
+&&       : {token, {'&&', TokenLine}}.
+\|\|       : {token, {'&&', TokenLine}}.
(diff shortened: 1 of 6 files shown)


src/alpaca_scan.xrl
```diff
@@ -131,10 +132,11 @@ c"(\\"*|\\.|[^"\\])*" :
 \+ : {token, {plus, TokenLine}}.
 
 
-
 [\*\/\%]   : {token, {int_math, TokenLine, TokenChars}}.
 {FLOAT_MATH} : {token, {float_math, TokenLine, TokenChars}}.
 ->       : {token, {'->', TokenLine}}.
+&&       : {token, {'&&', TokenLine}}.
+\|\|       : {token, {'&&', TokenLine}}.
```
j14159: Shouldn't this be `{token, {'||', TokenLine}}.`?

j14159:
Should `and` / `or` be infix?  I think that might be confusing as they don't use punctuation.  Maybe we want to treat them like functions e.g. `and false false`?

Licenser:
I just went by the erlang standards where both and `and` and `andalso` are infix opperators. I'm not sure having them as special functions seems odd to me. If they are a function they could be implemented with `&&` and `||` outside of the core alnguage.

Licenser:
As discussed on IRC, `and` and `or` are gone now as they could be implemented using `&&` and `||` if someone needed them

j14159:
Thanks for this!

Licenser:
:D wooohhh proud day a non trivial contribution!

PR #185, opened 2017-06-14
No warnings

Licenser:
Remove all warnings and add `warnings_as_errors` to make sure it stays that way ;)

### DIFF
--- a/src/alpaca_typer.erl
+++ b/src/alpaca_typer.erl
@@ -27,9 +27,6 @@
 
 -module(alpaca_typer).
 
--dialyzer({nowarn_function, dump_env/1}).
--dialyzer({nowarn_function, dump_term/1}).
-
 -include("alpaca_ast.hrl").
 -include("builtin_types.hrl").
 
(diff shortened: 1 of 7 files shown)


src/alpaca_typer.erl
```diff
@@ -2816,9 +2815,9 @@ typ_of(Env, Exp) ->
         {Typ, NewVar} -> {unwrap(Typ), update_counter(NewVar, Env)}
     end.
 
-from_code(C) ->
-    {ok, E} = alpaca_ast_gen:parse(alpaca_scanner:scan(C)),
-    E.
+%% from_code(C) ->
```
j14159: Meant to be removed?

Licenser: I didn't knew where it was coming from, it's not called but I figured it might have some reason. I'm happy to take it out so ;)
there are two more of those lingering around.

Licenser: dump_env to be precise

Licenser: and all gone ;)

j14159:
Thanks for this!

Licenser:
Always :)

PR #186, opened 2017-06-14
Remove purge warnings for nicer test output

Licenser:
Purge modules before deleting them during test runs. That turns this

image

into this

image

### DIFF
--- a/src/alpaca.erl
+++ b/src/alpaca.erl
@@ -169,6 +169,11 @@ type_modules(Mods) ->
 
 -ifdef(TEST).
 
+
+pd(Module) ->
+    code:purge(Module),
+    code:delete(Module).
+
 basic_file_test() ->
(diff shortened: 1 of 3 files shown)


Licenser:
The failing tests here is the following code:
```ml
module a
type vHX 'aDiPCKBIIl3Qk_MFBoA_ 'iZ 's3B_ 'k = A | B | C | D | { a : atom , b : atom , fn : atom } | E
type a = F
```

It fails with: `{error,{parse_error,test,1,{syntax_error,"fn"}}}`

I think that is a bug in the generator since `fn` is a reserved word and should not be a key?

j14159:
`fn` recently became a reserved word for lambdas, probably just needs a minor update to the ct suite.

Licenser:
fixed the proper tests too.

j14159:
Fixes #187 

Thanks for this!

PR #188, opened 2017-06-15
[WIP] Improve errors

Licenser:
Adding some details to errors.

**wip do not merge**

So far:
* Adding information about duplicate symbols (new is the location where the symbol was originailly defined):

image.

### DIFF
--- a/src/alpaca_error_format.erl
+++ b/src/alpaca_error_format.erl
@@ -19,6 +19,7 @@
 -ignore_xref([ fmt/2 ]).
 
 -compile({parse_transform, epo_gettext}).
+-include("alpaca_ast.hrl").
 
 %% number of lines to show before or after the errorrous line
 -define(CTX_AREA, 2).
@@ -59,6 +60,10 @@ fmt({error, Err}, Locale) ->
 ident(S) ->
     re:replace(S, "^", "  ", [multiline, global | ?RE_OPTS]).
 
+
+fmt_parse_error(#cannot_unify{path = Path, ast = Ast, expected = E, found =F}, _Locale) ->
+    {ok, R} = alpaca_walker:walk(Path, Ast),
+    cf("Error in: ~ts\nexpected a ~p but got a ~p", [alpaca_walker:fmt(R), E, F]);
 fmt_parse_error({duplicate_definition, Type, Id, Module, Line}, Locale) ->
     t(__(<<"duplicate_definition %(type) %(id) %(mod) %(line)">>),
       Locale, [{type, a2b(Type)}, {id, red(Id)}, {mod, file(a2b(Module))},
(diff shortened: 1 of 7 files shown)


Licenser:
I'm looking at typer errors those seem to be completely unhanlded so far. What I gathered so far is when a type error happend we get something like this:
`{cannot_unify, Module, Line, A, B}`

Where `Module` and `Line` are obvious, `A` is the 'acceptable' type and `B` is the provided one. And those can get as complex as they want.

I used the following core to create a errrps:
```ml
let add x y = x + y;;
let plus2 x = add x 2.0
```
Which will tell me that it expects an `t_int` but gets an `t_float`. I could make that an error.  Line:

```
modue:3
  Can't unify types, expected int but got float.
```

**However** what I'd like to do is a bit more. Something like:

```
module:3
  The second argument of the call to `add/2` `2.0` is a float. `add/2` defined in line 2 expects a float in this position.
```


This is a rather clean case so saying that in line X there is a `t_float` that should be a `t_int` will probably be enough for most people. Still it'd be really nice if we could point to `2.0` being the offending token, that would make things already a. But to achieve that we'd need some additional info. I think we'd need the AST for the offending type not only the type, as well as (optionally) the AST 1 level higher (to tell where in the code the wrong type is called).

I'm not sure how feasible that is as I think we'd need to start that quite low down in the whole compiling shebang but  I think it'd improve the whole experience by quite a bit.

Licenser:
Okay I've added some path tracking during the typer phase, it's far from perfect but it does give some. The failing tests is currently broken on purpose to get the output of it, it prints:

`Error in: add x 2.00\nexpected a t_int but got a t_float\n`

The text isn't great but it serves as a demo of what I'm aiming for and perhaps as a foundation of a discussion weather it's sensible or not.

What surprised me a bit is that the AST isn't type annotated by the typer, is that on purpose, impossible or is that just not done because it wasn't needed?

I'm also a bit unsure about how to extract the exact position of the error in apply, I could track the failing case down to that call `add x 2.00` but could not get it to tell me it's the 2.0 that's failing, a hint where to look for that kind of thing would be nice :)

src/alpaca_error_format.erl
```diff
@@ -408,10 +408,9 @@ double_binding_c_test() ->
         " h :: y -> h\n",
     Msg = comp_c(Source),
     Expected = <<"\e[0;36m\e[4m<no file>\e[0m:\e[0;36m3\e[0m\e[0m\n"
-                 "  Duplicate definition of binding "
-                 "\"\e[38;2;208;28;139my\e[0m\".\n"
-                 "  \"\e[38;2;208;28;139my\e[0m\" was priviously defined "
-                 "in \e[0;36m\e[4mtest_comp\e[0m:\e[0;36m\e[4m2\e[0m\n\e[0m">>,
+                 "  Duplicate definition of binding \"\e[0;31my\e[0m\".\n"
+                 "  \"\e[0;31my\e[0m\" was priviously defined in"
```
danabr: s/priviously/previously/

Licenser: thanks :)

j14159:
@Licenser terribly sorry it's taken me *this long* to get to your questions!

> What surprised me a bit is that the AST isn't type annotated by the typer, is that on purpose, impossible or is that just not done because it wasn't needed?

Initially I wanted the typer to perform as little mutation of the AST as possible but we actually do need to annotate the full graph in order to go further with match exhaustiveness checks, @danabr noted this a little while ago in his initial work on the problem.

> I'm also a bit unsure about how to extract the exact position of the error in apply, I could track the failing case down to that call add x 2.00 but could not get it to tell me it's the 2.0 that's failing, a hint where to look for that kind of thing would be nice :)

This is actually a little tricky and (I think) boils down to how function application is typed.  I suspect this is buried in `aplaca_typer:unify_list/4` since function application basically:

- makes a synthetic arrow out of the call, in this case the typer says "OK, `add x 2.00` should have the shape `fn A B -> C` where `B: float`"
- retrieves the _existing_ arrow from the environment, here `add: fn int int -> int`
- unifies the two lists of arguments, `[A, float]` and `[int, int]` (where the error pops up)
- unifies the return types, `C` and `int`

I'm simplifying a little here but I bet what's happening is the loss of detail in running through the unification of two lists.

I'll think on this some more, hope that helps but happy to bang this around here a bit more or in IRC.

PR #189, opened 2017-06-16
Revert back to base colors

Licenser:
After the discussion in #181 this reverts back to using base colours so they can be remapped as required by users.


j14159:
Thanks for this, and to @sanmiguel for the feedback!

Issue #133, opened 2017-02-08
Type specifications for top-level bindings


j14159:
I've been thinking of something relatively simple, e.g.

    val add: [int, int] -> int
    let add x y = x + y

    type option 'a = Some 'a | None

    val map 'a 'b: [option 'a, (['a] -> 'b)] -> option 'b
    let map Some a f = Some (f a)
    let map _ _ = None

We can attach the specs to binding defs, type the bindings themselves and then if a spec can unify with the inferred type, we use the provided spec.

For receivers maybe something like the following for a function that when spawned receives `'a`s and never terminates:

    val f 'a receives 'a: ['a] -> rec
    let f init_state =
      receive with
        next_state -> f next_state

None of the above addresses naming arguments for documentation purposes, e.g. in `map`'s case it would be good for a single docstring to be able to refer to each parameter with a single name to document the interface for users.

    val map 'a 'b: [option_in: option 'a, f: (['a] -> 'b] -> option 'b

The above reads poorly to me.  Something that should be left to docstrings themselves maybe?

danabr:
I experimented a bit with this a few weeks ago (using the syntax you suggest here). It worked OK for my needs, although you have to handle problems such as multiple val declarations for the same function.

Another option I played with is to put the type after the let binding. That would allow us to type any let binding at any level, not just top level function definitions:
````
let add x y : [int, int] -> int = x + y

let map Some a f : [option 'a, (['a -> 'b])]  -> option 'b
let map _ _ = None

type foo = int
let top_level_value : foo = 3
````
The downside is that you would have to inspect the type declaration itself to find out which type variables are used. Also, it might seem weird or inconsistent to only assign a type spec to one of the variants.

Finally, I've pondered more about the syntax of function type declarations. I think I've found the option I like the best is `fn int int -> int`. It is short and sweet, and you could reuse the `fn` keyword for lambdas. Reusing the examples above:
````
let add x y : fn int int -> int = x + y

let map Some a f : fn (option 'a)  (fn 'a -> 'b)  -> option 'b
let map _ _ = None
````

danabr:
And of course, a third option is to let you annotate function arguments and arbitrary expressions, yielding:
````
let add x:int y:int = x + y : int
let map (Some a) : (option 'a)  f : (fn 'a -> 'b) = Some (f a) : option 'b
let map _  _  = None
````
[...]

PR #190, opened 2017-07-05
Add type signatures

lepoetemaudit:
Adds type signatures that use the same style as type specs on ADTs, e.g.

```f#
val apply 'a 'b : fn (fn 'a -> 'b) 'a -> 'b
let apply f x = f x
```
This also matches the syntax suggested by @danabr in https://github.com/alpaca-lang/alpaca/issues/133. There are many examples provided in https://github.com/lepoetemaudit/alpaca/blob/a70f1fc051616a376da63c65bfb313f02e8da246/test_files/basic_type_signature.alp

Note that we separate function arguments only by whitespace, not by `->`, and preceded by `fn`, unlike other MLs. I believe this syntax was chosen to better reflect the fact that functions in Alpaca are multi-arity, not single argument and auto-curried (currying happens at call-site). However, as has been discussed it is possible we might support the more ML style syntax in future alongside or in place of this one.

### DIFF
--- a/Tour.md
+++ b/Tour.md
@@ -369,7 +369,7 @@ Most of the time, you can safely omit type signatures. They can be useful in get
 
 By forcing `print_string` to take only strings, we guard against passing types that Erlang's `io:put_chars` can't understand, preventing a runtime error.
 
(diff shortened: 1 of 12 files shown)


danabr:
This is so nice! I was experimenting with adding this a few months back, but never got it to work fully. I think this is a great addition. I especially had the FFI use case in mind back when I worked on this.

j14159:
Thanks for this, have really been looking forward to it!

yurrriq:
:tada:

PR #192, opened 2017-07-12
Simple tgz release building and publishing

j14159:
@lepoetemaudit does the way the tarball gets laid out match the expectations in your rebar3 work?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ script:
 - rebar3 eunit
 - rebar3 eunit
 - rebar3 ct
+- rebar3 as default compile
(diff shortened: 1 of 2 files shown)


lepoetemaudit:
@j14159 this looks perfect, thanks!

lepoetemaudit:
I just realised it collates the beams from the test profile; is that an issue?

j14159:
Good point - that'll have a lot of extra stuff included.  I'll take a look at making a different profile.

j14159:
OK, now once tests pass travis-ci will compile with the `default` profile and these will get bundled in the release instead.

PR #193, opened 2017-07-13
Don't check exhaustiveness on precompiled

lepoetemaudit:
When recompiling a module with dependencies, the exhaustiveness checker was also examining any precompiled modules. As the AST is stripped in precompiled modules, the checker was warning on every function definition, leading to a lot of unnecessary and confusing noise in the compiler output.


j14159:
Looks good, thanks for this!  With respect to the AST being stripped we might want to start looking at https://github.com/erlang/otp/pull/1367 a little more.

PR #199, opened 2017-07-19
Limit retyping of modules

lepoetemaudit:
The typer was retyping whole modules on every far reference it encountered, leading to performance issues. Also, the ordering of modules fed to the compiler, which the rebar plugin supplies in dependency order, was being reversed. This PR rectifies both of these problems.

I couldn't think of a hygienic way of knowing whether a module had been fully typed or not, so I added a flag to the `#alpaca_module{}` record. Happy to change this if there's a better way!

### DIFF
--- a/src/alpaca.erl
+++ b/src/alpaca.erl
@@ -136,7 +136,7 @@ maybe_print_exhaustivess_warnings(Warnings, Opts) ->
   end.
 
 load_files(Filenames) ->
-    lists:foldl(
+    Loaded = lists:foldl(
       fun(FN, Acc) ->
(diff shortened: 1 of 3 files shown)


lepoetemaudit:
Whoops, I broke things.

j14159:
Thanks for this!  I seem to recall be mystified before about the order of modules returned from compilation, nice to see it was something simple.  FWIW I think the `typed` flag is completely fine :)

lepoetemaudit:
Ah great, the typed flag certainly made it very simple! Thanks for the merge.

PR #201, opened 2017-07-19
Fixes #200, add lambda rewrite for clauses (match, FFI, receive)

### LINKED ISSUE #200: Lambdas defined inside match expressions cause crashes during code generation
Lambdas inside match expressions seem to trip something up in code generation. I believe it's another rewriting issue. An example of a failing module:

```fsharp
module bad

val fails : fn bool -> fn int -> int
let fails v =
    match v with
      | true -> fn y -> y * y
      | false -> fn y -> y + y
    
val succeeds : fn bool -> fn int -> int
let succeeds v =
    match v with
      | true -> (let f y = y * y in f)
      | false -> (let f y = y + y in f)
```

A stack trace of the resulting crash can be found here: https://gist.github.com/lepoetemaudit/43f41d8c8b753294b05694826ef738c3

The above example is contrived, but as far as I can tell a lambda declared anywhere on the right side of a match rule causes this issue.

lepoetemaudit:
Fixes #200; I realised this issue would affect all the expressions that use `#alpaca_clause{}`, i.e. `match`, `beam` and `receive`, so this fix covers all of these.

Thanks @j14159 for your pointers - they saved me quite a lot of time figuring this one out.

### DIFF
--- a/src/alpaca.erl
+++ b/src/alpaca.erl
@@ -669,6 +669,11 @@ use_lambda_test() ->
     [M] = compile_and_load(Files, [test]),
     ?assertMatch(15, M:useLambda(5)),
     ?assertMatch(13, M:useLambdaTuple(3)),
+    MatchFun = M:matchLambda(true),
+    Res = MatchFun(16),
+    ?assertMatch(256, Res),
+    FFIFun = M:ffiLambda("alpaca"),
+    ?assertMatch(64, FFIFun(8)),
     pd(M).
 
 -endif.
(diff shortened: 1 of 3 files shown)


j14159:
Glad I could help, thanks for fixing this!  Looks like the new methods in `test_files/use_lambda.alp` don't get exercised by tests yet - might be worth it?

lepoetemaudit:
Good point - I'll add.

lepoetemaudit:
So actually testing the returned result of the new additions was a good call - it turns out that the lambda being returned by `lambdaMatch` in the use_lambda.alp is always the one in the last clause, no matter whether true or false is passed into it! I'll investigate.

j14159:
Oh interesting!  Nice catch, thanks :)

lepoetemaudit:
I was able to recreate the odd lambda behaviour consistently with this test case;

```elm
module lambda
 
export main

-- apply two functions to `value`, return a tuple of each result
let apply2 f1 f2 value = 
  (f1 value, f2 value)

-- Erroneously returns (3, 3), i.e. the first lambda is overwritten by the second
let main () =
  apply2 (fn x -> x + 1) (fn x -> x + 2) 1
```

I fixed this by incrementing the synth lambda counter in what I think is the right place and added a test for this as well. Looks like this PR fixes two bugs for the price of one! Very glad I added the extra tests now.

j14159:
This is great, thanks for digging in!

PR #203, opened 2017-07-28
Allow expressions in list items

lepoetemaudit:
I noticed list items always required parens as it was requiring terms rather than expressions. This meant even simple literal lists such as `[1 + 2, 3 + 4]` wouldn't compile which isn't consistent with most langs.


j14159:
Nice one, thanks for this!

PR #204, opened 2017-07-31
Allow unsized binaries in construction

lepoetemaudit:
I noticed some odd behavior when constructing binaries:

```ocaml
let a = <<"hello ">> in
let b = <<"world">> in
<<a: type=binary, b: type=binary>>
```
...would yield `<<"hworld">>`. The compiler was (rightfully) disallowing unsized binary bindings in pattern matches, as Erlang does, but it was also preventing them from working when constructing binaries in the above manner where it's really useful to be able to concatenate binaries in this fashion.

This PR changes the approach a little bit, validating match clauses in the parser, and otherwise allowing non-tail position binary symbols to use the default size. This makes the behaviour consistent with Erlang's, and best of all means we can define a string concat operator directly in Alpaca:

```fsharp
val (^^) : fn string string -> string
let (^^) str1 str2 =
  match <<str1: type=utf8, str2: type=utf8>> with
  | <<res: type=utf8>> -> res

let example () =
  "hello " ^^ "world" ^^ "!!!"
```

### DIFF
--- a/src/alpaca_ast_gen.erl
+++ b/src/alpaca_ast_gen.erl
@@ -1764,7 +1764,10 @@ binary_test_() ->
                              value={'Symbol', #{line := 1, name := <<"a">>}}},
                           #alpaca_bits{
                              value={'Symbol', #{line := 1, name := <<"b">>}}}]}},
-        parse(alpaca_scanner:scan("<<a: size=8 type=int, b: size=8 type=int>>")))
+        parse(alpaca_scanner:scan("<<a: size=8 type=int, b: size=8 type=int>>"))),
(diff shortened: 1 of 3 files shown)


j14159:
Nice catch, thanks for this!  Really looking forward to seeing what you're building with Alpaca that's turning up all these bugs at some point :)

lepoetemaudit:
:) This particular one came from a port of Elm's HTML library I'm working on as part of the building blocks for a simple blog engine. Mostly though finding the Alpaca bugs and fixing them is the goal (and the fun part!)

PR #205, opened 2017-08-05
Add ^ xor logical operator

Licenser:


lepoetemaudit:
@Licenser I'm probably missing something, but as this is logical xor, doesn't it have the same behaviour as not equals (`!=`)? Ignoring my shell font which changes `!=` into a ligature version, this is the output I get from Alpaca:

image

Licenser:
It is, but it' might be less clear if you want to express logic.

lepoetemaudit:
That makes sense; I think part of my objection is that `^` isn't consistent with `&&` and `||` (they look like C's operators, and `^` is bitwise in C, not logical), so if we wanted to merge I'd have a preference for it to be expressed as `^^`.

@j14159 other than what I stated above I approve of the implementation in the PR; I think it'd better be your decision on whether you're happy to accept the redundancy for the sake of clarity.

Licenser:
I think the difference is that `&&` and `||` are short circuiting, while `^` isn't (and can't as it needs to know both sides of the operation to make up its mind. I did try the `^^` too but I've never seen it anywhere and it felt wrong that said I'm not hell bound on it.

j14159:
I'm unsure about consistency tbh, I feel like maybe logic operators should all follow the same format but I'm not 100% on this yet.

A question though, mostly curiosity:  type-checker implications aside, are there any performance concerns we should have with respect to implementing this as a match vs as a simple `!=` with the type of `^` or `^^` pre-set in the typer's environment?

Licenser:
If `^` is too confusing I can change it to `^^` my reasoning was that I've never seen `^^` anywhere as a `xor` while `^` at least in c-ish languages is a known operator.

Performance wise, I honestly am not sure, I went for this implementation was I wasn't sure how one would approach writing it as a not equal for booleans only and this is in line with the other boolean operators.

ferd:
afaict using C as a model is not the best since it conflates true/false booleans with 1 and 0, so some operators can be used in both cases without technically carrying well in other contexts. With this overloading, it makes sense for `&` to be AND, and for `|` to be OR, and by extension to use && and || as shortcircuit variants. XOR cannot have a shortcircuit variant, so `^^` has no context in which it makes sense historically or not.

- `^` is used in multiple programming languages, even though in set theory ⊻ or ⊕ are generally used
- other programming languages use an explicit `xor` operator (especially if you've got `rem` and `mod` already)
- apparently >< is sometimes used according to wikipedia (though I'd find it somewhat close to <> which has an entirely different meaning)

j14159:
I think @ferd makes some good points, obviously some conflicts around pattern matches and `|` though.

Uncomfortable question:  with the upcoming type annotations for v0.2.8, should we be maybe pulling all the logic operators out of the main language into something like `alpaca_lib`?  Less assumptions in the main language, more flexibility.

lepoetemaudit:
@j14159 it's tempting to put these operators into `alpaca_lib`; I tried defining `&&` and `||` infix (with different names) in the REPL and it worked fine, but I don't think we could implement the short-circuiting if we defined these as actual functions as all the parameters will have already been evaluated.

I think I'd rather keep them as pure expressions in Alpaca, but having them as functions would be neat. Elm has both - you can use (&&) and (||) as functions, but in the usual basic expression form it outputs as a Javascript expression so the short-circuit behaviour is retained. This issue has some more details:
https://github.com/elm-lang/core/issues/495

Licenser:
I think we had a discussion about a core language bootstrap of some alpaca functions before and it was not desired? For `||` and `&&` I think it's simply not possible w/o making the core language a lot more complex but I could be wrong, the short circuiting w/o macros or a way to not evaluate function arguments would just not happen.

I was playing with xor as a keyword but it felt kind of wrong given we use `&&` and `||`

j14159:
@Licenser @lepoetemaudit good points about the short-circuiting, I didn't think that through properly.

@lepoetemaudit interesting discussion in that Elm issue you linked.  I think I like having these operators available as functions but am definitely a little shy about it as we're privileging them with the short-circuiting stuff.  I believe I'm comfortable erring on the side of "FYI, these operators short-circuit and they're special - you can't do that with other functions" in the docs for now.

I do think @ferd is right to question which symbol we use for XOR though I'm not sure what we should use (`^` actually makes me think of exponent operators, I'm sure there are dozens of takes and opinions around this regardless :P )

Licenser:
Oh I only noticed the short circuting after trying to implement it that way ;) and running into the problem the first time around doing `and` and `or`.

Well we are early on enough that we can switch to `or`, `and` and `xor` those make it very clear what is happening and are definitely not ambiguous.

SML (which alpaca kind of reminds me of) uses `andalso` and `orelse` for short circuting, `and` and `or` for boolean a bit like Erlang.

It's hard that there are no standards -.-

j14159:
might be worth switching to `or`, `and`, `xor` as the defaults for clarity/simplicity tbh.  @Licenser how do you feel about that?

Any other thoughts/opinions?

Licenser:
I got no objections to `or`, `and` and `xor`, at least on my side `||` and `&&` depending on language flow it might even read nicer in the end (I don't have any large alpaca projects to check what'd feel more natural.)

lepoetemaudit:
I've had a change of heart and I think switching to `or`, `and`, `xor` is the way to go. I was a little wary initially as we've kept infix to symbols but I actually now think it's nice to disambiguate 'special' non-function infix like these logical operators.

There's no reason we couldn't also define `||` and `&&` as infix functions in the standard library, so users could treat them like functions, use them in higher order functions etc., but they would also explicitly not have short-circuit behaviour.

Licenser:
@j14159 I think it's just your verdict left :P and I'll update the PR accordingly ;)

j14159:
I'm good with `and`/`or`/`xor` :)  Thanks!

j14159:
Thanks for this @Licenser!

Question, seeking opinions:  do we want `and`/`or`/`xor` to be infix?  It feels like we're introducing an inconsistency.  Thoughts?

Licenser:
I think yes, as they are a special case, `and(A, B)` would read as if `A` and `B` are always evaluated, however `B` might not be. `A and B` makes this a lot clearer.

j14159:
@Licenser sorry to leave this open and waiting so long, been thinking a lot about this (and am currently out of town onboarding at a new job).  I'm quite torn between special-casing `and`/`or`/`xor` for infix use.  I think you make a good point about conveying the short-circuiting nature but on the other hand I'm also getting a little hung up on infix "functions" that look like all the other prefix functions rather than other infix functions :D  (the consistency concern)

I'm OK with merging this as-is, let's try it all out and see how it feels though we _might_ want to revisit prefix vs infix before v1.0.0 based on usage and more feedback.

Thanks for contributing all of this, I think it's going to be very useful!

PR #206, opened 2017-08-18
add erlang 20 for test

getong:


j14159:
Thanks for this!  Sorry to leave you hanging here, was on a brief vacation with no computer for a few days :)

PR #208, opened 2017-08-31
Type printing

lepoetemaudit:
This PR introduces a new module – `alpaca_printer` – with functions for returning textual (binary) representations of types. It works with `#alpaca_binding{}`, any arbitrary type supported in Alpaca (and user defined types like ADTs), and will also dump whole modules.

I believe this is more useful for tooling rather than by end-users (we can use it for Alpaca's own error messages and the shell / documentation generators can also make use of it) so the functions are only exposed from the `alpaca_printer` module itself. This will hopefully aid @Licenser with his work in improving errors in #188 

I believe the printer is pretty exhaustive - I've tried it against a few 'real' modules of my own, but types can be subtle and complex, especially with deep-nested types. Any missed cases should be easy to add, though.

I envisage that this module will also print docstrings once we have those.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -1,4 +1,4 @@
-{erl_opts, [debug_info, warnings_as_errors, nowarn_export_all, {gettext, alpaca_compiled_po}]}.
+{erl_opts, [debug_info, warnings_as_errors, {gettext, alpaca_compiled_po}]}.
 {xrl_opts, [{report, true}, {verbose, true}]}.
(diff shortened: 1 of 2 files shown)


j14159:
Looks great, thanks for this!

I wonder if we should maybe look at doing something with [PropEr](https://github.com/manopapad/proper) here along the lines of what @ypaq did for the parser.  Maybe we can produce source files with a wide spectrum of types, parse and format them, then _re_ -parse the signatures and match against the ones inferred from the generated ones.  Probably requires a full code formatter that uses this for type signatures though!  :)

PR #212, opened 2017-10-22
Reference cells in ETS instead of processes

j14159:
After a lot of false starts I ended up taking a lazy approach here.  I
suspect it might be more efficient to carry the ETS table around in
the environment but threading that through everything starts to get
even messier.


j14159:
This shaves about 0.2-0.3 seconds off the eunit test run as well.  Probably room for more gains here.

PR #220, opened 2018-01-02
Syntax highlight code snippets in README

lpil:
GitHub handily highlights code snippets in markdown files if annotated
with the language. Seeing as Alpaca is not yet supported I've annotated
them as Elm, which has a similar enough syntax.

We could also annotate files within the repo using a comment if desired.
The ReasonML folks did this before it was added to GitHub.


lpil:
Thanks!

PR #222, opened 2018-01-02
Correct README test example

lpil:
The test example in the README does not compile due to 2 undefined
variables. I've replaced it with one from ./test_files/


j14159:
Thanks for this!  I'll re-run the tests here and look at merging as soon as #225 is resolved.

j14159:
@lpil would you mind rebasing on master for the new build process?  Thanks!

Issue #225, opened 2018-01-03
Update Travis build to use release from github


j14159:
Now that v0.2.8 is tagged, `master` branch should build using the `master` branch of the rebar3 plugin and the v0.2.8 build in the releases section of this repository.

PR #226, opened 2018-01-03
Travis to fetch correct Alpaca release for building Alpaca

j14159:
Should now fetch the appropriate Alpaca release based on the current OTP version and builds with latest rebar3 plugin.


j14159:
Fixes #225

PR #239, opened 2018-01-12
Add syntax highlighting to Tour.md

lpil:


j14159:
Thanks for this.  Could you also please delete `Tour.org` in this PR?  Historically we've generated `Tour.md` with it but I'm not sure how syntax tags will work with this so we might as well just maintain one tour doc now.

lpil:
Updated :)

j14159:
Thanks for this!

